PR #1, opened 2021-11-02
Initial implementation of cluster-proxy addon

yue9944882:


controllers/clustermanagementaddon_controller.go
```diff
@@ -0,0 +1,590 @@
...
+									SecretName: config.Spec.Authentication.CertificateMounting.Secrets.SigningAgentServerSecretName,
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+	if err := c.Client.Create(context.TODO(), agentAuthServiceAccount); err != nil {
```
skeeey, line 280: We only create the resources, we may need to consider the update cases

controllers/clustermanagementaddon_controller.go
```diff
@@ -0,0 +1,590 @@
...
+		return reconcile.Result{}, nil
+	}
+
+	// get the related proxy configuration
+	config := &proxyv1alpha1.ManagedProxyConfiguration{}
+	if err := c.Client.Get(ctx, types.NamespacedName{
+		Name: addon.Spec.AddOnConfiguration.CRName,
+	}, config); err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Info("Cannot find proxy-configuration", "name", addon.Spec.AddOnConfiguration.CRName)
```
skeeey, line 97: We may conside the deletaion case, if there is no config, we may need to cleean up the deploy

pkg/addon/agent/agent.go
```diff
@@ -0,0 +1,276 @@
...
+		if entrySvc, err := p.nativeClient.CoreV1().
+			Services(config.Spec.ProxyServer.Namespace).
+			Get(context.TODO(), config.Spec.ProxyServer.Entrypoint.LoadBalancerService.Name, metav1.GetOptions{}); err == nil {
+			if len(entrySvc.Status.LoadBalancer.Ingress) > 0 {
+				lbEndpoint = entrySvc.Status.LoadBalancer.Ingress[0].IP
+			}
+		}
+	}
+	deploying := []runtime.Object{
+		newCASecret(addon.Spec.InstallNamespace, AgentCASecretName, p.selfSigner.CAData()),
```
skeeey, line 73: we may also need to deploy install namesapce

pkg/addon/agent/agent.go
```diff
@@ -0,0 +1,276 @@
...
+	AgentCASecretName = "cluster-proxy-ca"
+)
+
+func newAgentDeployment(clusterName, targetNamespace string, proxyConfig *proxyv1alpha1.ManagedProxyConfiguration, loadBalancerEndpoint string) *appsv1.Deployment {
+	serviceEntryPoint := proxyConfig.Spec.ProxyServer.InClusterServiceName + "." + proxyConfig.Spec.ProxyServer.Namespace
+	if len(proxyConfig.Spec.ProxyAgent.ProxyServerHost) > 0 {
+		serviceEntryPoint = proxyConfig.Spec.ProxyAgent.ProxyServerHost
+	} else if len(loadBalancerEndpoint) > 0 {
+		serviceEntryPoint = loadBalancerEndpoint
+	}
```
skeeey, line 169: we may need handle the case that if there is no `proxyConfig.Spec.ProxyAgent.ProxyServerHost` and the service is not lb,

skeeey: and could we get the `loadBalancerEndpoint` at here, I think it may be more clear, and it's better to record some log here

pkg/apis/proxy/v1alpha1/managedproxyconfiguration_types.go
```diff
@@ -0,0 +1,209 @@
...
+)
+
+func init() {
+	SchemeBuilder.Register(&ManagedProxyConfiguration{}, &ManagedProxyConfigurationList{})
+}
+
+// ManagedProxyConfigurationSpec defines the desired state of ManagedProxyConfiguration
+type ManagedProxyConfigurationSpec struct {
+	// +required
+	Authentication ManagedProxyConfigurationAuthentication `json:"authentication"`
```
skeeey, line 30: We may need some doc on these fields

pkg/addon/agent/agent.go
```diff
@@ -0,0 +1,276 @@
...
+						Subject: addonv1alpha1.Subject{
+							User: common.SubjectUserClusterProxyAgent,
+							Groups: []string{
+								common.SubjectGroupClusterProxy,
+							},
+						},
+					},
+				}
+			},
+			CSRApproveCheck: func(cluster *clusterv1.ManagedCluster, addon *addonv1alpha1.ManagedClusterAddOn, csr *csrv1.CertificateSigningRequest) bool {
```
skeeey, line 97: we may also need verify the csr

yue9944882:
@skeeey thx for reviewing! i will get back addressing the comments days later. am flooded by other tasks for the moment.

qiujian16:
I think we should merge this one at first and refresh it in the following. 

/approve
leave to @skeeey to tag.

xuezhaojun:
Found some "No newline at end of file" issue, may need changes before or after merge.

controllers/clustermanagementaddon_controller.go
yue9944882: now it handles update

controllers/clustermanagementaddon_controller.go
yue9944882: the deletion is working by linking the proxy config as owner to those provisioned resources i.e. these resources will be cleaned upon proxy config's deletion. wdyt?

pkg/addon/agent/agent.go
yue9944882: fixed

pkg/addon/agent/agent.go
yue9944882: clarified the logic of endpoint resolving

pkg/apis/proxy/v1alpha1/managedproxyconfiguration_types.go
yue9944882: yes i added api doc thru out the new go types. PTAL

pkg/addon/agent/agent.go
yue9944882: can you elaborate? i didnt see the proper way of verifying CSRs

yue9944882:
> Found some "No newline at end of file" issues, may need changes before or after merge.

can you name me the files? oddly i think it should be handled by go fmt tho.

xuezhaojun:
> > Found some "No newline at end of file" issues, may need changes before or after merge.
> 
> can you name me the files? oddly i think it should be handled by go fmt tho.

Sure, like `charts/cluster-proxy/templates/clusterrole.yaml`, maybe because it is yaml not go file.

pkg/apis/proxy/v1alpha1/managedproxyconfiguration_types.go
skeeey: It's good to me now

pkg/addon/agent/agent.go
skeeey: we need check the content of the csr to avoid the bad csr, there is an example:
https://github.com/open-cluster-management/cluster-proxy-addon/blob/main/pkg/hub/addon/addon.go#L159

skeeey:
> > > Found some "No newline at end of file" issues, may need changes before or after merge.
> > 
> > 
> > can you name me the files? oddly i think it should be handled by go fmt tho.
> 
> Sure, like `charts/cluster-proxy/templates/clusterrole.yaml`, maybe because it is yaml not go file.

Yes, most yaml files have no newline at end of file, and the `.gitignore` also has no newline

skeeey:
/lgtm

It's good to merge as the initial pr, we will continue to enhance this in the future

PR #9, opened 2021-11-24
Fix: log init

champly:
fix: Error log message not print when `--agent-image-name` not set.


yue9944882 (commented):
/lgtm
/approve

@champly thx!

PR #15, opened 2021-11-30
Bugfix: Entrypoint should be included in the SANs

yue9944882:
> Signed-off-by: yue9944882 <291271447@qq.com>


qiujian16 (commented):
/lgtm

PR #27, opened 2021-12-18
Feat: Adding a port-forward mode proxy entrypoint

yue9944882:


pkg/util/portforward.go
```diff
@@ -0,0 +1,221 @@
...
+		return err
+	}
+	defer streamConn.Close()
+
+	// create error stream
+	headers := http.Header{}
+	headers.Set(v1.StreamType, v1.StreamTypeError)
+	headers.Set(v1.PortHeader, fmt.Sprintf("%d", r.targetPort))
+	headers.Set(v1.PortForwardRequestIDHeader, strconv.Itoa(r.reqId))
+	r.reqId++
```
qiujian16, line 122: add a todo here also? I think you would need to at least lock here.

pkg/util/portforward.go
```diff
@@ -0,0 +1,221 @@
...
+		runtime.HandleError(fmt.Errorf("error creating forwarding stream for port %d -> %d: %v", r.targetPort, r.targetPort, err))
+		return err
+	}
+
+	localError := make(chan struct{})
+	remoteDone := make(chan struct{})
+
+	go func() {
+		// Copy from the remote side to the local port.
+		if _, err := io.Copy(conn, dataStream); err != nil && !strings.Contains(err.Error(), "use of closed network connection") {
```
qiujian16, line 156: why do we need to exclude the closed error?

pkg/util/portforward.go
yue9944882: added a lock

pkg/util/portforward.go
yue9944882: https://github.com/kubernetes/kubernetes/blob/cd6ffff85d257ff9067d59339f2ffdbcc66dc164/staging/src/k8s.io/client-go/tools/portforward/portforward.go#L309-L310

actually this line is copied from the original client-go's port-forward server. for now the client-go library cannot be reused by cluster-proxy due to some limitation

qiujian16 (commented):
/lgtm

PR #29, opened 2021-12-20
Support set entrypoint lb annotations

champly:
Signed-off-by: champly <champly@outlook.com>

In some situation, lb which have internal ip for secrity, will set lb annotation, such as:
``` yaml
service.beta.kubernetes.io/alibaba-cloud-loadbalancer-address-type: "intranet" // for alibaba cloud.
service.beta.kubernetes.io/azure-load-balancer-internal: true // for azure cloud.
cloud.google.com/load-balancer-type: "Internal" // for google cloud.
```

### DIFF
--- a/controllers/clustermanagementaddon_controller.go
+++ b/controllers/clustermanagementaddon_controller.go
@@ -273,7 +273,7 @@ func (c *ClusterManagementAddonReconciler) ensureEntrypoint(config *proxyv1alpha
 			ObjectMeta: metav1.ObjectMeta{
 				Namespace:   config.Spec.ProxyServer.Namespace,
 				Name:        config.Spec.ProxyServer.Entrypoint.LoadBalancerService.Name,
-				Annotations: config.Spec.ProxyServer.Entrypoint.LoadBalancerService.Annotations,
+				Annotations: getAnnotation(config.Spec.ProxyServer.Entrypoint.LoadBalancerService.Annotations),
 			},
 			Spec: corev1.ServiceSpec{
 				Selector: map[string]string{
@@ -487,4 +487,22 @@ func getPEMCertExpireTime(pemBytes []byte) *metav1.Time {
 		return nil
 	}
 	return &metav1.Time{Time: cert.NotAfter}
+}
+
+func getAnnotation(list []proxyv1alpha1.AnnotationVar) map[string]string {
+	if len(list) == 0 {
+		return nil
+	}
+	annotation := make(map[string]string, len(list))
+	for _, v := range list {
+		key := v.Key
+		if len(key) == 0 {
+			continue
+		}
+		if len(key) > 63 {
+			key = key[:63]
+		}
+		annotation[key] = v.Value
+	}
+	return annotation
 }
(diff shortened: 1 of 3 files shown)


pkg/apis/proxy/v1alpha1/managedproxyconfiguration_types.go
```diff
@@ -244,6 +244,11 @@ type EntryPointLoadBalancerService struct {
 	// +optional
 	// +kubebuilder:default=proxy-agent-entrypoint
 	Name string `json:"name"`
+
+	// annoations is the annoations of the load-balancer service.
+	// if the service use internal-ip, should have some annotations.
+	// +optional
+	Annotations map[string]string `json:"annotations"`
```
yue9944882, line 251: it's overall not recommended to define a map field in the custom resource. consider making this a key-ed list such as:
```
myData
- key: foo
   value: bar
```

see another example at: https://github.com/open-cluster-management-io/api/blob/8b5ab9358f31451f4debc8b578021b217c2cce39/work/v1/types.go#L300-L303

yue9944882: cc @qiujian16 about the api convention

pkg/apis/proxy/v1alpha1/managedproxyconfiguration_types.go
```diff
@@ -244,6 +244,11 @@ type EntryPointLoadBalancerService struct {
 	// +optional
 	// +kubebuilder:default=proxy-agent-entrypoint
 	Name string `json:"name"`
+
+	// annoations is the annoations of the load-balancer service.
+	// if the service use internal-ip, should have some annotations.
```
yue9944882, line 249:
```
// Annotations is the annoations of the load-balancer service. This is for allowing customizing service using vendor-specific extended annotations such as:
// - service.beta.kubernetes.io/alibaba-cloud-loadbalancer-address-type: "intranet"
// - service.beta.kubernetes.io/azure-load-balancer-internal: true
```

pkg/apis/proxy/v1alpha1/managedproxyconfiguration_types.go
yue9944882, line 251: nit: omitempty

yue9944882 (commented):
@champly thanks, i will send another pr to make the addon-manager reconcile the service annotation upon configuration changes in a follow-up.

pkg/apis/proxy/v1alpha1/managedproxyconfiguration_types.go
champly: > it's overall not recommended to define a map field in the custom resource. consider making this a key-ed list such as:
> 
> ```
> myData
> - key: foo
>    value: bar
> ```
> 
> see another example at: https://github.com/open-cluster-management-io/api/blob/8b5ab9358f31451f4debc8b578021b217c2cce39/work/v1/types.go#L300-L303

modify with sugesttion. PTAL

controllers/clustermanagementaddon_controller.go
```diff
@@ -487,3 +488,21 @@ func getPEMCertExpireTime(pemBytes []byte) *metav1.Time {
...
+	if len(list) == 0 {
+		return nil
+	}
+	annotation := make(map[string]string, len(list))
+	for _, v := range list {
+		key := v.Key
+		if len(key) == 0 {
+			continue
+		}
+		if len(key) > 63 {
```
yue9944882, line 502: either conform the label validation w/ the k8s original library i.e. `validation.IsQualifiedName` on the key and `validation.IsValidLabelValue` on the value or removed the validation to make it simpler

yue9944882 (commented):
thanks, one nit otherwise LGTM

champly:
> thanks, one nit otherwise LGTM

Modify with validation.

yue9944882 (commented):
/lgtm
/approve

PR #43, opened 2022-01-04
E2E: Adding more cases

yue9944882:


qiujian16 (commented):
/lgtm

thanks, I think you would need to add it into action as the followup

yue9944882:
@qiujian16 it's already in the action for clarification

PR #49, opened 2022-01-12
Refresh managedclusteraddon status's last observed generation

yue9944882:


controllers/clustermanagementaddon_controller.go
```diff
@@ -152,43 +153,52 @@ func (c *ClusterManagementAddonReconciler) Reconcile(ctx context.Context, reques
 	}
 
 	// deploying central proxy server instances into the hub cluster.
-	err = c.deployProxyServer(config)
+	isRedeployed, err := c.deployProxyServer(config)
```
qiujian16, line 156: maybe modified or deployed. isRedeployed is bit confusing

controllers/clustermanagementaddon_controller.go
```diff
@@ -197,55 +207,84 @@ func (c *ClusterManagementAddonReconciler) deployProxyServer(config *proxyv1alph
 		newProxyServerRole(config),
 		newProxyServerRoleBinding(config),
 	}
+	deployed := false
 	for _, resource := range resources {
-		if err := c.ensure(resource); err != nil {
-			return err
+		created, updated, err := c.ensure(config.Generation, resource)
```
qiujian16, line 212: ditto. I think it is not necessary to diff between created/updated.

controllers/clustermanagementaddon_controller.go
```diff
@@ -152,43 +153,52 @@ func (c *ClusterManagementAddonReconciler) Reconcile(ctx context.Context, reques
...
 	}
 	now := metav1.Now()
 	editingConfig := config.DeepCopy()
-	editingConfig.Status = status
+	editingConfig.Status = expectingStatus
 	for i := range editingConfig.Status.Conditions {
-		editingConfig.Status.Conditions[i].LastTransitionTime = now
+		expectingCondition := editingConfig.Status.Conditions[i]
+		currentCondition := meta.FindStatusCondition(config.Status.Conditions, expectingCondition.Type)
+		if currentCondition == nil || currentCondition.Status != expectingCondition.Status {
```
qiujian16, line 192: maybe use meta.SetStatusCondition here

yue9944882: great!

controllers/clustermanagementaddon_controller.go
yue9944882: renamed to isModified

controllers/clustermanagementaddon_controller.go
yue9944882: actually i clarified them on purpose, to record events when any resources are modified. this will be helpful for us to find out the timeline when some resources are unexpectedly updated.

yue9944882:
@qiujian16 let's move to #50, rebasing the other pull onto this one has too many conflicts..

PR #50, opened 2022-01-12
E2E: healthcheck

yue9944882:


controllers/clustermanagementaddon_controller.go
```diff
@@ -197,55 +207,84 @@ func (c *ClusterManagementAddonReconciler) deployProxyServer(config *proxyv1alph
...
+		return created, false, nil
+	}
+
+	// update if generation bumped
+	if !created && int(incomingGeneration) > currentGeneration {
+		annotations := resource.GetAnnotations()
+		if annotations == nil {
+			annotations = make(map[string]string)
+		}
+		annotations[common.AnnotationKeyConfigurationGeneration] = strconv.Itoa(int(incomingGeneration))
```
qiujian16, line 275: I think you use this to track if a resource should be updated upon update of clustermanamanagementaddon? We might need to comment on this.

controllers/clustermanagementaddon_controller.go
```diff
@@ -152,43 +153,52 @@ func (c *ClusterManagementAddonReconciler) Reconcile(ctx context.Context, reques
...
 		return nil
 	}
 	now := metav1.Now()
 	editingConfig := config.DeepCopy()
-	editingConfig.Status = status
+	editingConfig.Status = expectingStatus
 	for i := range editingConfig.Status.Conditions {
-		editingConfig.Status.Conditions[i].LastTransitionTime = now
+		expectingCondition := editingConfig.Status.Conditions[i]
+		currentCondition := meta.FindStatusCondition(config.Status.Conditions, expectingCondition.Type)
```
qiujian16, line 191: I think you can use meta.SetStatusCondition here which will handle transition time.

controllers/clustermanagementaddon_controller.go
```diff
@@ -197,55 +207,84 @@ func (c *ClusterManagementAddonReconciler) deployProxyServer(config *proxyv1alph
 		newProxyServerRole(config),
 		newProxyServerRoleBinding(config),
 	}
+	deployed := false
 	for _, resource := range resources {
-		if err := c.ensure(resource); err != nil {
-			return err
+		created, updated, err := c.ensure(config.Generation, resource)
```
qiujian16, line 212: nit I might only consider if the resource is modified (created or updated). It seems created, updated can be reduced to 1 var

controllers/clustermanagementaddon_controller.go
yue9944882: updated

controllers/clustermanagementaddon_controller.go
yue9944882: right, added comments to clarify the usage of the generation label

controllers/clustermanagementaddon_controller.go
yue9944882: actually i clarified them on purpose, to record events when any resources are modified. this will be helpful for us to find out the timeline when some resources are unexpectedly updated.

qiujian16 (commented):
/lgtm

PR #66, opened 2022-01-28
Refactor: Deploy agent-addon via helm templates

yue9944882:


qiujian16:
FYI @zhiweiyin318 you may want to take a look :)

pkg/proxyagent/agent/agent.go
```diff
@@ -126,9 +65,89 @@ func (p *proxyAgent) GetAgentAddonOptions() agent.AgentAddonOptions {
 			CSRApproveCheck: func(cluster *clusterv1.ManagedCluster, addon *addonv1alpha1.ManagedClusterAddOn, csr *csrv1.CertificateSigningRequest) bool {
 				return cluster.Spec.HubAcceptsClient
 			},
-			PermissionConfig: p.setupPermission,
+			PermissionConfig: NewClusterProxySetupPermissionFunc(runtimeClient, nativeClient),
 			CSRSign:          CustomSignerWithExpiry(ProxyAgentSignerName, caKeyData, caCertData, time.Hour*24*180),
-		},
+		}).
+		WithInstallStrategy(agent.InstallAllStrategy(common.AddonInstallNamespace)).
+		WithScheme(scheme).
```
zhiweiyin318, line 72: we could ignore `WithScheme` if we have no customized resources. anyway it is ok to configure.

pkg/proxyagent/agent/manifests/charts/addon-agent/templates/namespace.yaml
```diff
@@ -0,0 +1,8 @@
+{{ if .Values.includeNamespaceCreation }}
```
yue9944882, line 1: another issue i found when integrating w/ the helm agent is that we have to deal w/ namespace creation explicitly in the template. conventionally in the work api, we inline the namespace object, however in a normal helm chart, namespaces are normally implicitly created using `install --create-namespace`. for now, i had an additional value `includeNamespaceCreation` which is disable upon manual chart install but enabled when it's converted to work. @zhiweiyin318 @qiujian16

pkg/proxyagent/agent/agent.go
yue9944882: removed

PR #70, opened 2022-02-16
Fix proxy-agent image to  proxyConfig.Spec.ProxyAgent.Image

xuezhaojun:

The inconsistent point right now is:

For proxy-server we use `config.Spec.ProxyServer.Image` as it's image.

https://github.com/open-cluster-management-io/cluster-proxy/blob/58c7d9318303cc47b6a9cd58779685e539df5430/pkg/proxyserver/controllers/manifests.go#L114

But for proxy-agent we are using the same image as addon-agent is using.

https://github.com/open-cluster-management-io/cluster-proxy/blob/58c7d9318303cc47b6a9cd58779685e539df5430/pkg/proxyagent/agent/manifests/charts/addon-agent/templates/addon-agent-deployment.yaml#L34

The PR is to allow users to use entire different anp(proxy-server & proxy-agent) images.

### DIFF
--- a/charts/cluster-proxy/values.yaml
+++ b/charts/cluster-proxy/values.yaml
@@ -12,8 +12,8 @@ replicas: 1
 
 spokeAddonNamespace: "open-cluster-management-cluster-proxy"
 
-proxyServerImage: cluster-proxy
-proxyAgentImage: cluster-proxy
+proxyServerImage: quay.io/open-cluster-management/cluster-proxy:v0.1.4
+proxyAgentImage: quay.io/open-cluster-management/cluster-proxy:v0.1.4
 
 proxyServer:
   entrypointLoadBalancer: false
--- a/charts/cluster-proxy/templates/managedproxyconfiguration.yaml
+++ b/charts/cluster-proxy/templates/managedproxyconfiguration.yaml
@@ -9,7 +9,7 @@ spec:
     signer:
       type: SelfSigned
   proxyServer:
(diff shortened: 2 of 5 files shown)


charts/cluster-proxy/values.yaml
```diff
@@ -12,8 +12,8 @@ replicas: 1
 
 spokeAddonNamespace: "open-cluster-management-cluster-proxy"
 
-proxyServerImage: cluster-proxy
-proxyAgentImage: cluster-proxy
+proxyServerImage: quay.io/open-cluster-management/cluster-proxy:v0.1.4
```
yue9944882, line 15: how about we merge the image values into one?

xuezhaojun: Sure we can do it. Do we need to update the API structure as well?

xuezhaojun: Old:
```
proxyServerImage: quay.io/open-cluster-management/cluster-proxy:v0.1.4
proxyAgentImage: quay.io/open-cluster-management/cluster-proxy:v0.1.4
```

New:
```
ANPImage: quay.io/open-cluster-management/cluster-proxy:v0.1.4
```
When we say "merge into one", Do we mean modification as the above?

yue9944882: oh the point of this pull is to separate the addon image and ANP's image, i was talking about converging all the images into one above. nvmd, lets get this in.

yue9944882 (commented):
/lgtm
/approve

Issue #74, opened 2022-02-24
Proxy-Agent support customize `proxy-server-port`


xuezhaojun:
I found the `proxy-agent` can set a customize port, so it use `8091` in default.

In my case,  I'm using `type: Hostname` as entrypoint type, and also using `443` port to expose the proxy-server.

So could we add a flag to support user to specify the `--proxy-server-port` in `proxy-agent`?

PS: 
It's a simple issue, and I think the only thing we should have a discussion about is the flags' name.
Should it be as the following?

```yaml
    proxyServer:
    image: {{ .Values.proxyServerImage }}
    namespace: {{ .Release.Namespace }}
    entrypoint:
      {{- if .Values.proxyServer.entrypointAddress }}
      type: Hostname
      hostname:
        value: {{ .Values.proxyServer.entrypointAddress }}
        port: {{ ..Values.proxyServer.entrypointPort }} # Here!! Updated
      {{- else if .Values.entrypointLoadBalancer }}
      type: LoadBalancerService
      loadBalancerService: {}
      {{- else  }}
      type: PortForward
      {{- end }}
```

agent.go add a flag: 
```go
serviceEntryPointPort = proxyConfig.Spec.ProxyServer.Entrypoint.Hostname.Port
```

I also have free time to fix this issue. 😁

PR #75, opened 2022-02-24
Add proxy-server-port in proxy-agent

xuezhaojun:

related issue: https://github.com/open-cluster-management-io/cluster-proxy/issues/74

### DIFF
--- a/pkg/apis/proxy/v1alpha1/managedproxyconfiguration_types.go
+++ b/pkg/apis/proxy/v1alpha1/managedproxyconfiguration_types.go
@@ -219,6 +219,12 @@ type ManagedProxyConfigurationProxyServerEntrypoint struct {
 	// `hostname` points to a fixed hostname for serving agents' handshakes.
 	// +optional
 	Hostname *EntryPointHostname `json:"hostname"`
+
+	// `port` is the target port to access proxy servers
+	// +optional
+	// +kubebuilder:default=8091
+	// +kubebuilder:validation:Minimum=1
+	Port int32 `json:"port,omitempty"`
 }
 
 // EntryPointType is the type of the entrypoint.
--- a/pkg/proxyagent/agent/agent.go
+++ b/pkg/proxyagent/agent/agent.go
@@ -156,6 +156,11 @@ func GetClusterProxyValueFunc(runtimeClient client.Client, nativeClient kubernet
 			annotations[common.AnnotationKeyConfigurationGeneration] = strconv.Itoa(int(proxyConfig.Generation))
 		}
 
+		serviceEntryPointPort := proxyConfig.Spec.ProxyServer.Entrypoint.Port
+		if serviceEntryPointPort == 0 {
+			serviceEntryPointPort = 8091
+		}
+
 		registry, image, tag := config.GetParsedAgentImage()
 		return map[string]interface{}{
 			"agentDeploymentName":      "cluster-proxy-proxy-agent",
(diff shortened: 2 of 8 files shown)


pkg/apis/proxy/v1alpha1/managedproxyconfiguration_types.go
```diff
@@ -219,6 +219,11 @@ type ManagedProxyConfigurationProxyServerEntrypoint struct {
 	// `hostname` points to a fixed hostname for serving agents' handshakes.
 	// +optional
 	Hostname *EntryPointHostname `json:"hostname"`
+
+	// `port` points to the port proxy-agent should access
+	// +optional
+	// +kubebuilder:default=8091
```
yue9944882, line 225: to keep the backward-compatbility, lets set the default to 443, also add a validation marker that asserts positive value

pkg/proxyagent/agent/agent.go
```diff
@@ -156,6 +156,8 @@ func GetClusterProxyValueFunc(runtimeClient client.Client, nativeClient kubernet
 			annotations[common.AnnotationKeyConfigurationGeneration] = strconv.Itoa(int(proxyConfig.Generation))
 		}
 
+		serviceEntryPointPort := proxyConfig.Spec.ProxyServer.Entrypoint.Port
```
yue9944882, line 159: also for data-plane compatibility, reset it to 443 if it's zero

pkg/apis/proxy/v1alpha1/managedproxyconfiguration_types.go
```diff
@@ -219,6 +219,11 @@ type ManagedProxyConfigurationProxyServerEntrypoint struct {
 	// `hostname` points to a fixed hostname for serving agents' handshakes.
 	// +optional
 	Hostname *EntryPointHostname `json:"hostname"`
+
+	// `port` points to the port proxy-agent should access
```
yue9944882, line 223: `port` is the target port to access proxy servers

pkg/apis/proxy/v1alpha1/managedproxyconfiguration_types.go
xuezhaojun: The thing is we need to set it to 8091 to keep the backward compatibility because ANP uses 8091 as the default agent-port. If we set the default value to 443, the new version could not access to the previous one.

https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/d2eec1ec3ac33f683c81a950f171caab568c0250/cmd/server/app/options/options.go#L308

pkg/proxyagent/agent/agent.go
xuezhaojun: Sure. Fixed.

pkg/apis/proxy/v1alpha1/managedproxyconfiguration_types.go
xuezhaojun: Comment udpated.

pkg/apis/proxy/v1alpha1/managedproxyconfiguration_types.go
```diff
@@ -219,6 +219,12 @@ type ManagedProxyConfigurationProxyServerEntrypoint struct {
 	// `hostname` points to a fixed hostname for serving agents' handshakes.
 	// +optional
 	Hostname *EntryPointHostname `json:"hostname"`
+
+	// `port`` is the target port to access proxy servers
```
yue9944882, line 223: nit: double black slash `

yue9944882 (commented):
/lgtm
/approve

Issue #71, opened 2022-02-17
Lack of some clusterrole


xuezhaojun:
When I try to install cluster-proxy on a OCP platform, I got the following error:

```
E0217 08:54:41.554025       1 base_controller.go:270] "cluster-management-addon-controller" controller failed to sync "local-cluster/cluster-proxy", err: managedclusteraddons.addon.open-cluster-management.io "cluster-proxy" is forbidden: cannot set an ownerRef on a resource you can't delete: , <nil>

E0217 08:54:41.944396       1 base_controller.go:270] "addon-registration-controller" controller failed to sync "local-cluster/cluster-proxy", err: roles.rbac.authorization.k8s.io "cluster-proxy-addon-agent" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>
```

I didn't find this issue in a KinD env, so I suppose OCP has a tighter permission on roles.

In my practice the issue can be solved by:

On hub:
Apply the following changes to `clusterrole.yaml`:
```yaml
  - apiGroups:
      - addon.open-cluster-management.io
    resources:
      - clustermanagementaddons
      - managedclusteraddons
      - clustermanagementaddons/status
      - managedclusteraddons/status
    verbs:
      - get
      - list
      - watch
      - create
      - update
      - patch
      - delete # !!! Here! add this
  - apiGroups:
      - proxy.open-cluster-management.io
    resources:
      - managedproxyconfigurations
      - managedproxyconfigurations/status
      - managedproxyconfigurations/finalizers # !!! Here! add this
    verbs:
      - get
      - list
      - watch
      - update
      - patch
```

One agent:

I'm not sure how to fix this issue on the agent,  because as the code shows:

https://github.com/open-cluster-management-io/cluster-proxy/blob/ffd6cad3bc8be95ec20bd8635ffab54b9e6e8678/pkg/proxyagent/agent/agent.go#L190-L195

We can either add permission to the addon-framework or we can delete the ownerReference here(Or set the BlockOwnerDeletion to false).

So it depends on whether this `OwnerReferences` must be in agent manifest work. @yue9944882 

cc @skeeey

yue9944882:
[...]

PR #80, opened 2022-03-07
Fix ownerreference bug caused by addon-framework

xuezhaojun:

This PR is related to: https://github.com/open-cluster-management-io/cluster-proxy/issues/71

But this ownerReference bug is not caused by cluster-proxy itself, but by addon-framework:

https://github.com/open-cluster-management-io/addon-framework/blob/f3917ca3a39cd63ede2bfbf63988c463c21ded2b/pkg/addonmanager/controllers/clustermanagement/controller.go#L111

In addon-framework,  we set `clustermanagementAddon` as ownerReference to our addon. So this permission also need added in the clusterrole.

### DIFF
--- a/charts/cluster-proxy/templates/clusterrole.yaml
+++ b/charts/cluster-proxy/templates/clusterrole.yaml
@@ -17,6 +17,7 @@ rules:
       - clustermanagementaddons
       - managedclusteraddons
       - clustermanagementaddons/status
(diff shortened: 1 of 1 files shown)


yue9944882 (commented):
/lgtm
/approve

yue9944882 (commented):
/lgtm
/approve

PR #85, opened 2022-03-22
Fix: wait for mgr started, then run test cases

xuezhaojun:


xuezhaojun:
The `mgr.Start` function is running in a goroutine, therefore sometimes mgr is not started when test cases are running and cause unit-test failure such as: https://github.com/open-cluster-management-io/cluster-proxy/runs/5549300121?check_suite_focus=true

xuezhaojun:
We found that by setting the `timeout` to 1 sec, we can produce this fail every time.
So the guess is, we can extend the `timeout` to reduce the possibility of fail.

PR #95, opened 2022-04-07
support set installnamespace on spoke

xuezhaojun:

### DIFF
--- a/pkg/apis/proxy/v1alpha1/managedproxyconfiguration_types.go
+++ b/pkg/apis/proxy/v1alpha1/managedproxyconfiguration_types.go
@@ -311,6 +311,10 @@ type ManagedProxyConfigurationProxyAgent struct {
 	// `additionalArgs` defines args used in proxy-agent.
 	// +optional
 	AdditionalArgs []string `json:"additionalArgs,omitempty"`
+	// `installNamespace` defines the namespace where the proxy-agent will be installed.
+	// +optional
+	// +default=open-cluster-management-cluster-proxy
+	InstallNamespace string `json:"installNamespace,omitempty"`
 }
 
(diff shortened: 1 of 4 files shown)


pkg/proxyagent/agent/agent.go
```diff
@@ -64,6 +64,24 @@ func NewAgentAddon(signer selfsigned.SelfSigner, signerNamespace string, v1CSRSu
 			},
 		})
 	}
+
+	// Get InstallNamespace of AddonAgent
+	clusterAddon := &addonv1alpha1.ClusterManagementAddOn{}
```
xuezhaojun, line 69: We can not move get here, because runtimeClient is not start yet.

pkg/apis/proxy/v1alpha1/managedproxyconfiguration_types.go
```diff
@@ -311,6 +311,10 @@ type ManagedProxyConfigurationProxyAgent struct {
 	// `additionalArgs` defines args used in proxy-agent.
 	// +optional
 	AdditionalArgs []string `json:"additionalArgs,omitempty"`
+	// `installNamespace` defines the namespace where the proxy-agent will be installed.
+	// +optional
+	// +default=open-cluster-management-cluster-proxy
+	InstallNamespace string `json:"installNamespace,omitempty"`
```
yue9944882, line 317: will the existing installed `ManagedClusterAddon` reinstall if we edit this value? or rather, is this a valid use-case? as far as i know i think the installed namespace should be a "static" configuration --- we dont need to edit this on the flight. so a `--install-namespace` flag for addon-manager will be sufficient for customizing the namespace, wdyt?

Issue #91, opened 2022-04-01
Change the usage of `leaseUpdater`


xuezhaojun:
Right now we are using hubConfigClient to new a leaseUpdater:

https://github.com/open-cluster-management-io/cluster-proxy/blob/dad7ce6c9f80aee1d01f21230b8f9fca7a9a16bf/cmd/addon-agent/main.go#L51

https://github.com/open-cluster-management-io/cluster-proxy/blob/dad7ce6c9f80aee1d01f21230b8f9fca7a9a16bf/pkg/proxyagent/health/lease.go#L10-L16

But a better usage is using `spokeClient` in `NewLeaseUpdater ` and using `hubConfigClient` in `WithHubLeaseConfig ` as the following code shows here:

https://github.com/stolostron/multicloud-operators-foundation/blob/b01ead83c195e3d43a8b34f2a1f0510c4698694d/cmd/agent/agent.go#L174

xuezhaojun:
> @xuezhaojun the PR was merged, I think we can close this now.

Sure, Thanks for the effort!

/Close

PR #101, opened 2022-04-18
Use spoke client to update lease

zhujian7:
Signed-off-by: zhujian <jiazhu@redhat.com>
issue reference: https://github.com/open-cluster-management-io/cluster-proxy/issues/91


pkg/proxyagent/agent/manifests/charts/addon-agent/templates/addon-agent-role.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
xuezhaojun, line 1: Adding role and rolebinding in manifest work may dup with https://github.com/open-cluster-management-io/cluster-proxy/blob/b8a5476fb3162cdec8bf85c4515fdd40b645a94f/pkg/proxyagent/agent/agent.go#L75

zhujian7: I think it's not a dup, since [cluster-proxy/pkg/proxyagent/agent/agent.go](https://github.com/open-cluster-management-io/cluster-proxy/blob/b8a5476fb3162cdec8bf85c4515fdd40b645a94f/pkg/proxyagent/agent/agent.go#L75-L102) is setting permission for the **HUB** cluster instead of the managed cluster.

xuezhaojun: I got it. You're right about this.

yue9944882 (commented):
/approve

sry for the delay

Issue #53, opened 2022-01-19
Support access not only kube-apiserver but also other services in the managed cluster


xuezhaojun:
Is it possible for cluster-proxy to support the above feature?

Currently, proxy-agent would direct all traffic into kube-apiserver through this ExternalName type service:

https://github.com/open-cluster-management-io/cluster-proxy/blob/c83c9c28fbd0ae10b751013233ad592a77f0bc48/pkg/proxyagent/agent/agent.go#L447-L463

But if we do the following:
1. Add the target service' hostname in proxy-agent  "agent-identifiers" flag. It can make sure requests will go to the desired managed cluster through the tunnel.
2. Add another ExternalName service to point to the target service.

Therefore another target service on the managed cluster could be accessed by the user on the hub as well.

showeimer:
Hi @yue994488!  Would you be able to evaluate the technical feasability of this enhancement request, use case, and how much work it may be?

xuezhaojun:
I have a draft idea about this, we can add a "map" struct `services` in `managedproxyconfigurations`:

```yaml
apiVersion: proxy.open-cluster-management.io/v1alpha1
kind: ManagedProxyConfiguration
metadata:
  name: cluster-proxy
spec:
  services: # Here!! This is what we need to add.
    search: search.namespace.svc
    ui: ui.namespace.svc
  authentication:
    dump:
      secrets: {}
    signer:
      type: SelfSigned
  proxyServer:
    image: quay.io/open-cluster-managment.io/cluster-proxy:latest
    ...
```
The key represents the service we want to access, and the value is the respective service address.

The agent will generate several ExternalName Services based on this map structure.

And also register on proxy-agent with more agent-identifiers:
* cluster1-search
* cluster1-ui

And user can use `cluster1-search` as the hostname of the request to access the search service on the agent(managed cluster).

Is this making sense? Please share more advice, thanks!

qiujian16:
FYI @yue9944882 this is case that we discussed today.

qiujian16:
@xuezhaojun I do not think it could be a cluster scoped configuration. I wonder this should be a per cluster or per clusterset configuration. Also it is not quite formal to use `cluster1-search` as the hostname. It probably could be `search.default.cluster.cluster1`

xuezhaojun:
Agree

yue9944882:
[...]

PR #128, opened 2022-09-14
Feature: Add ServiceResolvers to expose services from managed clusters to a hub.

xuezhaojun:

Related issue: https://github.com/open-cluster-management-io/cluster-proxy/issues/53


xuezhaojun:
Working on adding e2e test cases...

xuezhaojun:
/assign @qiujian16 
/assign @yue9944882 

Although the e2e test cases are still on working, I guess we can review API part at first.

pkg/apis/proxy/v1alpha1/managedproxyserviceresolver_types.go
```diff
@@ -0,0 +1,70 @@
...
+
+// ManagedProxyServiceResolverList contains a list of ManagedProxyServiceResolver
+type ManagedProxyServiceResolverList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []ManagedProxyServiceResolver `json:"items"`
+}
+
+// ManagedProxyServiceResolverSpec defines the desired state of ManagedProxyServiceResolver.
+type ManagedProxyServiceResolverSpec struct {
```
qiujian16, line 52: it is not clear on how a consumer can access the service.

pkg/util/serviceurl.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"crypto/sha256"
+	"fmt"
+)
+
+func GenerateServiceURL(cluster, namespace, service string) string {
+	// Using hash to generate a random string; URL string should be less than 63 characters
+	// Keep namespace and service for debugging purpose
+	return fmt.Sprintf("%s-%s-%x", namespace, service, sha256.Sum256([]byte(fmt.Sprintf("%s_%s_%s", cluster, namespace, service))))[:(63 - len(namespace) - len(service) - 1)]
```
qiujian16, line 11: what does `:(63 - len(namespace) - len(service) - 1)` mean? how do you ensure it is positive?

pkg/proxyagent/agent/manifests/charts/addon-agent/templates/addon-agent-deployment.yaml
```diff
@@ -38,7 +38,7 @@ spec:
           args:
             - --proxy-server-host={{ .Values.serviceEntryPoint }}
             - --proxy-server-port={{ .Values.serviceEntryPointPort }}
-            - --agent-identifiers={{ list (printf "host=%s" .Values.clusterName) (printf "host=%s.%s" .Values.clusterName .Release.Namespace) (printf "host=%s.%s.%s" .Values.clusterName .Release.Namespace .Values.serviceDomain) | join "&" }}
+            - --agent-identifiers={{ .Values.agentIdentifiers }}
```
qiujian16, line 41: is this scalable, what if we have 100 services?

pkg/proxyagent/agent/agent.go
```diff
@@ -235,3 +255,41 @@ const (
...
 	AgentCASecretName = "cluster-proxy-ca"
 )
+
+func getAgentIndendifiersAndOtherServices(clusterName string, clusterlabels map[string]string, serviceResolvers []proxyv1alpha1.ManagedProxyServiceResolver, runtimeClient client.Client) (string, []map[string]string, error) {
+	var aids []string
+	var otherServices []map[string]string
+
+	aids = append(aids, fmt.Sprintf("host=%s", clusterName))
+	aids = append(aids, fmt.Sprintf("host=%s.%s", clusterName, config.AddonInstallNamespace))
+	aids = append(aids, fmt.Sprintf("host=%s.%s.%s", clusterName, config.AddonInstallNamespace, "svc.cluster.local"))
```
qiujian16, line 265: why svc.cluster.local? a k8s cluster may not use it as postfix.

pkg/proxyserver/operator/eventhandler/triggerallclustermanagementaddon_event.go
```diff
@@ -0,0 +1,51 @@
...
+
+// findClusterProxyAddon will triger clustermanagementaddon on all managed clusters to reconcile.
+func (m *TriggerAllClusterManagementAddonsHandler) findClusterProxyAddon(limitingInterface workqueue.RateLimitingInterface) {
+	list := &addonv1alpha1.ClusterManagementAddOnList{}
+	err := m.Client.List(context.TODO(), list)
+	if err != nil {
+		return
+	}
+	for _, addon := range list.Items {
+		if addon.Spec.AddOnConfiguration.CRDName == crdName {
```
qiujian16, line 44: this is being deperecated. I think when Resolver is changed, you will always need to update the certain addon, right? it does not need to check the crd name. for cluster set, it is a bit different. You would also need to check if the clusterset is in a certain resolver.

pkg/util/serviceurl.go
xuezhaojun: Yeah, that's a bug, it should be a fixed prefix.

xuezhaojun: Updated.

pkg/proxyagent/agent/manifests/charts/addon-agent/templates/addon-agent-deployment.yaml
xuezhaojun: It should be as ANP doesn't mention any limitation on the number of `agent-identifiers`.
But it also still worth to add a test case for this, I will note this down.

pkg/proxyagent/agent/agent.go
xuezhaojun: This is copy and paste from the previous code:
```
--agent-identifiers={{ list (printf "host=%s" .Values.clusterName) (printf "host=%s.%s" .Values.clusterName .Release.Namespace) (printf "host=%s.%s.%s" .Values.clusterName .Release.Namespace .Values.serviceDomain) | join "&" }}
```

The `serviceDomain` is hard code using `svc.cluster.local`. But actually, we are not using this host.

If @yue9944882 is also not using it, I guess we can remove this.

pkg/apis/proxy/v1alpha1/managedproxyserviceresolver_types.go
xuezhaojun: Updated the description and add the usage example for consumers.

pkg/proxyserver/operator/eventhandler/triggerallclustermanagementaddon_event.go
xuezhaojun: A little confused here, "it does not need to check the crd name."
If we don't check the CRD name, is there any way else for us to know if it's a "cluster-proxy" addon?

xuezhaojun: Split a event handler for `clusterset` having match check before insert into queue.

pkg/apis/proxy/v1alpha1/managedproxyserviceresolver_types.go
```diff
@@ -0,0 +1,70 @@
...
+
+	// ServiceName represents the name of the service.
+	// +required
+	ServiceName string `json:"serviceName"`
+}
+
+// ManagedProxyServiceResolverStatus defines the observed state of ManagedProxyServiceResolver.
+type ManagedProxyServiceResolverStatus struct {
+	// Conditions contains the different condition statuses for this ManagedProxyServiceResolver.
+	Conditions []metav1.Condition `json:"conditions"`
```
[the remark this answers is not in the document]
xuezhaojun: Right now, there is only one condition to represent the availability of the managedclusterset in the spec section.

pkg/proxyserver/operator/eventhandler/triggerallclustermanagementaddon_event.go
qiujian16: you can use the name of the addon.

xuezhaojun: Do we mean this field? https://github.com/open-cluster-management-io/cluster-proxy/blob/641a5dbc71eed83423abb36960c931972f1ce9c7/charts/cluster-proxy/templates/clustermanagementaddon.yaml#L4

Is there any possibility that a customer only use our image and a customized chart with this field doesn't equals to "cluster-proxy"?

qiujian16: will it work in addon-framework? I think the addon-name is set in addon-framework

xuezhaojun: Got it. Also thanks to @skeeey for the explanation. 
Updated all filter functions to using the name instead of CRDName.

xuezhaojun:
The port-forward works fine locally but break in github action env. Working on it.

yue9944882:
generally LGTM, plz squash the commits before we get this in, thanks

xuezhaojun:
> generally LGTM, plz squash the commits before we get this in, thanks

Thanks for the review, the commits is squashed and all

pkg/proxyagent/agent/agent.go
xuezhaojun: Check the doc and saw the format is: `my-svc.my-namespace.svc.cluster-domain.example`.
May be we can keep it, in case someone want to use a hostname with '.cluster.local'?

xuezhaojun:
Hello, @qiujian16 PTAL, thanks!

xuezhaojun:
@qiujian16  please take a review of this API change:

```
type ManagedProxyServiceResolver struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`

	Spec   ManagedProxyServiceResolverSpec   `json:"spec,omitempty"`
	Status ManagedProxyServiceResolverStatus `json:"status,omitempty"`
}

//+kubebuilder:object:root=true

// ManagedProxyServiceResolverList contains a list of ManagedProxyServiceResolver
type ManagedProxyServiceResolverList struct {
	metav1.TypeMeta `json:",inline"`
	metav1.ListMeta `json:"metadata,omitempty"`
	Items           []ManagedProxyServiceResolver `json:"items"`
}

// ManagedProxyServiceResolverSpec defines the desired state of ManagedProxyServiceResolver.
type ManagedProxyServiceResolverSpec struct {
	// ManagedClusterSelector selects a set of managed clusters.
	// +required
	ManagedClusterSelector ManagedClusterSelector `json:"managedClusterSelector"`

	// ServiceSelector selects a service.
	// +required
	ServiceSelector ServiceSelector `json:"serviceSelector"`
}

// ManagedClusterSelectorType is the type of ManagedClusterSelector.
// +kubebuilder:validation:Enum=ManagedClusterSet
type ManagedClusterSelectorType string

var (
	// ManagedClusterSetSelectorType indicates the selector is a ManagedClusterSet.
	// In this type, the manageclusterset field of the selector is required.
	ManagedClusterSelectorTypeClusterSet ManagedClusterSelectorType = "ManagedClusterSet"
)

type ManagedClusterSelector struct {
	// Type represents the type of the selector.
	// +required
	ManagedClusterSelectorType ManagedClusterSelectorType `json:"type"`

	// ManagedClusterSet defines a set of managed clusters that need to expose the service.
	// +optional
	ManagedClusterSet string `json:"managedClusterSet,omitempty"`
}

// ServiceSelectorType is the type of ServiceSelector.
// +kubebuilder:validation:Enum=NamespacedServiceName
type ServiceSelectorType string

var (
	// ServiceSelectorTypeLabel indicates the selector is a label selector.
	// In this type, the serviceNamespace and serviceName fields of the selector is required.
	ServiceSelectorTypeNamespacedServiceName ServiceSelectorType = "NamespacedServiceName"
)

type ServiceSelector struct {
	// Type represents the type of the selector.
	// +required
	ServiceSelectorType ServiceSelectorType `json:"type"`

	// ServiceNamespace represents the namespace of the service.
	// +optional
	ServiceNamespace string `json:"serviceNamespace,omitempty"`

	// ServiceName represents the name of the service.
	// +optional
	ServiceName string `json:"serviceName,omitempty"`
}

```

pkg/proxyagent/agent/agent.go
```diff
@@ -194,6 +199,68 @@ func GetClusterProxyValueFunc(
...
+		serviceResolvers := filterLegalManagedProxyServiceResolver(serviceResolverList.Items)
+
+		// deleted serviceResolvers are not included in the list
+		serviceResolvers = filterManagedProxyServiceResolverByDeletionTimestamp(serviceResolvers)
+
+		// filter serviceResolvers by managedClusterSet
+		serviceResolvers = filterManagedProxyServiceResolverWithManagedClusterSet(serviceResolvers, managedClusterSetMap)
+
+		// convert serviceResolvers to otherServices
+		otherServices := []map[string]string{}
```
qiujian16, line 245: s/otherServices/servicesToExpose/

pkg/proxyagent/agent/agent.go
```diff
@@ -194,6 +199,67 @@ func GetClusterProxyValueFunc(
...
+		otherServices = removeDupAndSortOtherServices(otherServices)
+
+		// add default agentIdentifiers
+		var aids []string
+		aids = append(aids, fmt.Sprintf("host=%s", cluster.Name))
+		aids = append(aids, fmt.Sprintf("host=%s.%s", cluster.Name, config.AddonInstallNamespace))
+
+		// add otherServices into aids
+		for _, otherService := range otherServices {
+			aids = append(aids, fmt.Sprintf("host=%s", otherService["url"]))
```
qiujian16, line 259: if we only care about url here, why we need otherServices as a []map[string]string?
can we make it sets.String, so you will not have duplicates easily

pkg/apis/proxy/v1alpha1/managedproxyserviceresolver_types.go
```diff
@@ -0,0 +1,131 @@
...
+
+var (
+	// ManagedClusterSetSelectorType indicates the selector is a ManagedClusterSet.
+	// In this type, the manageclusterset field of the selector is required.
+	ManagedClusterSelectorTypeClusterSet ManagedClusterSelectorType = "ManagedClusterSet"
+)
+
+type ManagedClusterSelector struct {
+	// Type represents the type of the selector. Now only ManagedClusterSet is supported.
+	// +required
```
qiujian16, line 77: enum and default marker

pkg/apis/proxy/v1alpha1/managedproxyserviceresolver_types.go
```diff
@@ -0,0 +1,131 @@
...
+
+var (
+	// ServiceSelectorTypeServiceRef indicates the selector requires serviceNamespace and serviceName fields.
+	ServiceSelectorTypeServiceRef ServiceSelectorType = "ServiceRef"
+)
+
+type ServiceSelector struct {
+	// Type represents the type of the selector. Now only ServiceRef type is supported.
+	// +required
+	ServiceSelectorType ServiceSelectorType `json:"type"`
```
qiujian16, line 104: enum and default marker

pkg/proxyagent/agent/agent.go
xuezhaojun: Done.

pkg/proxyagent/agent/agent.go
xuezhaojun: Both `urls` and `otherServices(now servicesToExpose)` is used to render manifest template.
URL is used in args `--agentIndentifiers` in agent deployments.
`servicesToExpose` is used in `pkg/proxyagent/agent/manifests/charts/addon-agent/templates/other-services.yaml` to generate `ExternalName` type services.

xuezhaojun: can we make it sets.String, so you will not have duplicates easily.

I'm not sure because `servicesToExpose` is finally used to render a manifest template, and an `array` makes this part easy.

pkg/apis/proxy/v1alpha1/managedproxyserviceresolver_types.go
xuezhaojun: Done.

pkg/apis/proxy/v1alpha1/managedproxyserviceresolver_types.go
xuezhaojun: Done.

pkg/proxyagent/agent/agent.go
```diff
@@ -194,6 +199,67 @@ func GetClusterProxyValueFunc(
...
+		managedClusterSets := filterManagedClusterSetByDeletionTimestamp(managedClusterSetList.Items)
+
+		// only cluseterSet cover current cluster include in the list.
+		managedClusterSets, err = filterManagedClusterSetForCurrentCluster(managedClusterSets, cluster.Labels)
+		if err != nil {
+			return nil, err
+		}
+
+		managedClusterSetMap := map[string]clusterv1beta1.ManagedClusterSet{}
+		for _, managedClusterSet := range managedClusterSets {
```
skeeey, line 224: could we use one loop to handle this?

```
for _,  managedClusterSets := range managedClusterSetList.Items {
		if !managedClusterSet.DeletionTimestamp.IsZero() {
			continue
		}
                selector, err := clusterv1beta1.BuildClusterSelector(&managedClusterSet)
		if err != nil {
			return nil, err
		}
		if !selector.Matches(labels.Set(clusterlabels)) {
			continue
		}

               managedClusterSetMap[managedClusterSet.Name] = managedClusterSet
               
}
```

pkg/proxyagent/agent/agent.go
```diff
@@ -194,6 +199,67 @@ func GetClusterProxyValueFunc(
...
+		// filter serviceResolvers by managedClusterSet
+		serviceResolvers = filterManagedProxyServiceResolverWithManagedClusterSet(serviceResolvers, managedClusterSetMap)
+
+		// convert serviceResolvers to servicesToExpose
+		servicesToExpose := []map[string]string{}
+		for _, serviceResolver := range serviceResolvers {
+			servicesToExpose = append(servicesToExpose, convertManagedProxyServiceResolverToService(cluster.Name, serviceResolver))
+		}
+		// remove dup and sort servicesToExpose
+		servicesToExpose = removeDupAndSortServices(servicesToExpose)
```
skeeey, line 250: could we use one loop to handle this and use set instead of []map[string]string{}?

```
servicesToExpose := k8s.io/apimachinery/pkg/util/sets.String{}
for _, serviceResolver := range serviceResolverList.Items {
		if !serviceResolver.DeletionTimestamp.IsZero() {
			continue
		}
                if !util.IsServiceResolverLegal(&serviceResolver) {
			continue
		}
               if _, ok := managedClusterSetMap[serviceResolver.Spec.ManagedClusterSelector.ManagedClusterSet.Name]; !ok {
			continue
		}

                servicesToExpose.Insert(util.GenerateServiceURL(clusterName, serviceResolver.Spec.ServiceSelector.ServiceRef.Namespace, serviceResolver.Spec.ServiceSelector.ServiceRef.Name))
} 

// add servicesToExpose into aids
for _, s := range servicesToExpose.List() {
     aids = append(aids, fmt.Sprintf("host=%s", s["url"]))
}

```

pkg/proxyagent/agent/agent.go
xuezhaojun: Yes, we can. but here is a trade-off for unit-test sake.

pkg/proxyagent/agent/agent.go
xuezhaojun: Also, like the above mentioned, use different functions to make unit-test cases clearer.

And we need a **sorted** services list here, and `sets.String` can't promise this.

pkg/proxyagent/agent/agent.go
skeeey: you can put this loop in one function, I think it is still easy to test with unit

pkg/proxyagent/agent/agent.go
skeeey: I think  the `aids ` and `servicesToExpose` is sorted here

why we need to sort the `servicesToExpose`? for test? 

the problem of `sets.String` is we cannot render the `Service` helm chart easily, right?
and I thought this agin, could we use `map[string]ManagedProxyServiceResolver` to render the `Service` helm chart  instead of map[string]string?

```
{{- range $key, $val := .Values.servicesToExpose }}
apiVersion: v1
kind: Service
metadata:
  namespace: {{ $.Release.Namespace }}
  name: {{ $key }}
spec:
  type: ExternalName
  externalName: {{ $val.Spec.ServiceSelector.ServiceRef.Name}}.{{ $val.Spec.ServiceSelector.ServiceRef.Namespace}}.svc.cluster.local
---
{{- end }}
```

or consider the `ServiceSelector` may has other type, we may make the map[string]string more simple
map[string]string{
    "url": "",
   "externalName": ""
}

examples/access-other-services/main.go
```diff
@@ -0,0 +1,106 @@
+package main
```
skeeey, line 1: `access-exported-services` ?

pkg/proxyserver/controllers/service_resolver_controller.go
```diff
@@ -0,0 +1,143 @@
...
+		Complete(c)
+}
+
+func (c *ServiceResolverReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log.Info("Reconciling ServiceResolver", "name", req.Name)
+	// refreshing service resolvers status
+	if err := c.refreshManageProxyServiceResolversStatus(); err != nil {
+		return reconcile.Result{}, err
+	}
+	return ctrl.Result{}, nil
```
skeeey, line 61: `return reconcile.Result{}, c.refreshManageProxyServiceResolversStatus()`

pkg/proxyagent/agent/agent.go
xuezhaojun: The render part won't be an issue for sure, the case needs sort is when (s1, s2, s3) turns to (s2, s3, s1), we don't want agent to re-deploy the manifests.

pkg/proxyserver/controllers/service_resolver_controller.go
xuezhaojun: Fixed.

pkg/proxyagent/agent/agent.go
xuezhaojun: Done.

pkg/proxyagent/agent/agent.go
xuezhaojun: Update.

pkg/proxyagent/agent/agent.go
```diff
@@ -231,7 +294,83 @@ func CustomSignerWithExpiry(customSignerName string, caKey, caData []byte, durat
...
 const (
 	ApiserverNetworkProxyLabelAddon = "open-cluster-management.io/addon"
-
-	AgentSecretName   = "cluster-proxy-open-cluster-management.io-proxy-agent-signer-client-cert"
-	AgentCASecretName = "cluster-proxy-ca"
+	AgentSecretName                 = "cluster-proxy-open-cluster-management.io-proxy-agent-signer-client-cert"
+	AgentCASecretName               = "cluster-proxy-ca"
 )
+
+func filterMCS(managedClusterSets []clusterv1beta1.ManagedClusterSet, clusterlabels map[string]string) ([]clusterv1beta1.ManagedClusterSet, error) {
```
skeeey, line 301: how about return the map[string]clusterv1beta1.ManagedClusterSet directly

pkg/proxyagent/agent/agent.go
```diff
@@ -231,7 +294,83 @@ func CustomSignerWithExpiry(customSignerName string, caKey, caData []byte, durat
...
+		if !selector.Matches(labels.Set(clusterlabels)) {
+			continue
+		}
+
+		filteredManagedClusterSets = append(filteredManagedClusterSets, mcs)
+	}
+	return filteredManagedClusterSets, nil
+}
+
+func filterMPSR(serviceResolvers []proxyv1alpha1.ManagedProxyServiceResolver, managedClusterSetMap map[string]clusterv1beta1.ManagedClusterSet) []proxyv1alpha1.ManagedProxyServiceResolver {
```
skeeey, line 325: how about return the serviceToExposes directly

pkg/proxyserver/controllers/service_resolver_controller.go
```diff
@@ -0,0 +1,131 @@
...
+						Type:   proxyv1alpha1.ConditionTypeServiceResolverAvaliable,
+						Status: metav1.ConditionTrue,
+						Reason: "ManagedProxyServiceResolverAvaliable",
+					}
+				}
+			}
+		}
+
+		// update status; now only consider one condition.
+		if len(currentConditions) != 0 {
```
skeeey, line 116: you may need the meta.FindStatusCondition

pkg/proxyagent/agent/agent.go
xuezhaojun: Done.

pkg/proxyserver/controllers/service_resolver_controller.go
xuezhaojun: Updated.

pkg/proxyagent/agent/agent.go
xuezhaojun: Updated, saved one turn of iter.

PR #130, opened 2022-09-22
Remove waitsycn function in application code because controller-runti…

xuezhaojun:
…me alredy did internally.

https://github.com/kubernetes-sigs/controller-runtime/blob/c83076e9f79258f23bcb200cf2391cf4ade939cf/pkg/internal/controller/controller.go#L208

So this part of code is dup with the framework code.


qiujian16 (commented):
/approve
/lgtm

PR #133, opened 2022-10-11
Fix agent template field and e2e test cases issue.

xuezhaojun:

### DIFF
--- a/pkg/proxyagent/agent/agent.go
+++ b/pkg/proxyagent/agent/agent.go
@@ -38,6 +38,11 @@ var FS embed.FS
 
 const (
 	ProxyAgentSignerName = "open-cluster-management.io/proxy-agent-signer"
+
+	// serviceDomain must added because coreDNS is not recursive resolver, and go dns client assume it is.
+	// See more details: https://coredns.io/manual/setups/#recursive-resolver; https://github.com/golang/go/blob/6f445a9db55f65e55c5be29d3c506ecf3be37915/src/net/dnsclient_unix.go#L666
+	// TODO: the serviceDomain should be configurable.
+	serviceDomain = "svc.cluster.local"
 )
 
 func NewAgentAddon(signer selfsigned.SelfSigner, signerNamespace string, v1CSRSupported bool, runtimeClient client.Client, nativeClient kubernetes.Interface) (agent.AgentAddon, error) {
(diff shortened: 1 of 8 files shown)


pkg/proxyagent/agent/agent.go
```diff
@@ -337,7 +337,7 @@ func managedProxyServiceResolverToFilterServiceToExpose(serviceResolvers []proxy
 func convertManagedProxyServiceResolverToService(clusterName string, sr proxyv1alpha1.ManagedProxyServiceResolver) serviceToExpose {
 	return serviceToExpose{
 		Host:         util.GenerateServiceURL(clusterName, sr.Spec.ServiceSelector.ServiceRef.Namespace, sr.Spec.ServiceSelector.ServiceRef.Name),
-		ExternalName: fmt.Sprintf("%s.%s.svc", sr.Spec.ServiceSelector.ServiceRef.Name, sr.Spec.ServiceSelector.ServiceRef.Namespace),
+		ExternalName: fmt.Sprintf("%s.%s.svc.cluster.local", sr.Spec.ServiceSelector.ServiceRef.Name, sr.Spec.ServiceSelector.ServiceRef.Namespace),
```
qiujian16, line 340: why cluster.local?

xuezhaojun: From the test, it seems that this a necessary suffix. But I need to make a deep invest to answer why... I will bring the result to this thread later.

xuezhaojun: Found in coreDNS document, says it's not a recursive resolver: https://coredns.io/manual/setups/#recursive-resolver

xuezhaojun: But in go dns client, it assumes that dns servers in resolv.conf should be a recursive-resolver.  https://github.com/golang/go/blob/6f445a9db55f65e55c5be29d3c506ecf3be37915/src/net/dnsclient_unix.go#L666

xuezhaojun:
@qiujian16 The todo comment is added. Another TODO about security and insecureSkip will be added in cluster-proxy-addon PR.

temp/tls.crt
```diff
@@ -0,0 +1,44 @@
+-----BEGIN CERTIFICATE-----
+MIIEEzCCAvugAwIBAgIIRPA1JEQTgLkwDQYJKoZIhvcNAQELBQAwMDEuMCwGA1UE
```
qiujian16, line 2: nit: we commonly do not put cert/key in github repo. Could them be autogenerated during test?

qiujian16 (commented):
/approve

just nit comment, we should not contain cert/key in the PR

temp/tls.crt
xuezhaojun: Sorry, they are miss upload test files. Deleted.

PR #135, opened 2022-10-16
Add getServiceDomain function for cases domain is customized.

xuezhaojun:

### DIFF
--- a/pkg/proxyagent/agent/agent.go
+++ b/pkg/proxyagent/agent/agent.go
@@ -5,6 +5,7 @@ import (
 	"embed"
 	"encoding/base64"
 	"fmt"
+	"net"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


pkg/proxyagent/agent/agent.go
```diff
@@ -370,3 +366,21 @@ func removeDupAndSortServices(services []serviceToExpose) []serviceToExpose {
...
 	return newServices
 }
+
+// The serviceDomain must added because coreDNS is not recursive resolver, and go dns client assume it is.
+// See more details: https://coredns.io/manual/setups/#recursive-resolver; https://github.com/golang/go/blob/6f445a9db55f65e55c5be29d3c506ecf3be37915/src/net/dnsclient_unix.go#L666
+func getServiceDomain() string {
+	apiSvc := "kubernetes.default"
+
+	// In normal case, it will return: `kubernetes.default.svc.cluster.local.`
+	cname, err := net.LookupCNAME(apiSvc)
```
qiujian16, line 376: I thin this is run on hub?

xuezhaojun: Yes, I forgot this part... we can run this function on a spoke and send the value back to hub. But I guess in that way it would be too complex...

xuezhaojun: Is it possible that we can use clusterlable or status to related this value to a managedcluster CR?

qiujian16: who on the spoke will check this?

xuezhaojun: It feels like a meta data of a managed cluster for me, so maybe the klusterlet?

PR #143, opened 2022-11-14
Fix: shoud update agent generation in all entrypoint types.

xuezhaojun:


xuezhaojun:
/assign @yue9944882 
It looks like annotation `proxy.open-cluster-management.io/configuration-generation` is not an `Entrypoint.Type` related value. So move it out of the switch.

yue9944882 (commented):
/lgtm
/approve

PR #147, opened 2022-12-15
fix the addon ns orphan

zhiweiyin318:
Signed-off-by: Zhiwei Yin <zyin@redhat.com>


pkg/proxyagent/agent/manifests/charts/addon-agent/templates/namespace.yaml
```diff
@@ -4,5 +4,5 @@ kind: Namespace
 metadata:
   name: {{ .Release.Namespace }}
   annotations:
-    helm.sh/resource-policy: keep
```
xuezhaojun, line 7: Note：this line doesn't actually work since the chart is deployed by addon-framework.

PR #148, opened 2023-01-03
Fix: add a flag to control whether we need the default kube-apiserver proxy.

xuezhaojun:


xuezhaojun:
/assign @qiujian16 
Hello, all checks look good, please take a review when you have free time, thanks!

qiujian16 (commented):
should we also update some args in the deployment of the agent?

xuezhaojun:
> should we also update some args in the deployment of the agent?

In this case, we don't need to, because the deployment of the agent is actually an ANP component and we can not modify it directly.

pkg/proxyagent/agent/agent.go
```diff
@@ -263,10 +266,13 @@ func GetClusterProxyValueFunc(
 
 		servicesToExpose := removeDupAndSortServices(managedProxyServiceResolverToFilterServiceToExpose(serviceResolverList.Items, managedClusterSetMap, cluster.Name))
 
-		// add default agentIdentifiers
 		var aids []string
-		aids = append(aids, fmt.Sprintf("host=%s", cluster.Name))
-		aids = append(aids, fmt.Sprintf("host=%s.%s", cluster.Name, config.AddonInstallNamespace))
+
+		// add default kube-apiserver agentIdentifiers
+		if enableKubeApiProxy {
```
xuezhaojun, line 272: @qiujian16 Here, we only add kube-apiserver as `agentIndentifiers` when enableKubeApiProxy equals `true`.

pkg/proxyagent/agent/manifests/charts/addon-agent/templates/cluster-service.yaml
```diff
@@ -1,3 +1,4 @@
+{{ if .Values.enableKubeApiProxy }}
```
xuezhaojun, line 1: The service is using managed cluster's name as its name, when managed cluster's name starts with a number, it will cause trouble because klusterlet-work-agent can not apply a Service with the name starts with a number. 

```
E1223 09:27:20.700245 1 base_controller.go:270] "ManifestWorkAgent" controller failed to sync "addon-cluster-proxy-deploy-0", err: Service "20pi8725abdjf5lccjlt61dnqidp1bm2" is invalid: metadata.name: Invalid value: "20pi8725abdjf5lccjlt61dnqidp1bm2": a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name', or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')
```

The newly added flag `enableKubeApiProxy` can remove this service from the Manifestwork when the flag is `false`.

PR #150, opened 2023-03-03
Init Codecov.

xuezhaojun:


.github/workflows/go-presubmit.yml
```diff
@@ -26,9 +23,6 @@ jobs:
     steps:
       - name: checkout code
         uses: actions/checkout@v2
-        with:
-          fetch-depth: 1
-          path: go/src/open-cluster-management.io/cluster-proxy
```
qiujian16, line 31: why we need to remove this field?

xuezhaojun: For 2 reasons:
1. Simplicity, it doesn't harm anything if we remove it.
2. Codecov needs `cover.out` to have the same file path as files in the GitHub repo. (We can config Codecov but it's more simple to remove the above thing)

PR #151, opened 2023-03-09
Fix: requeue when managedproxyconfiguration not found.

xuezhaojun:
This PR is to fix the issue when the first reconcile failed because `managedproxyconfiguration` not found.

### DIFF
--- a/pkg/proxyserver/controllers/clustermanagementaddon_controller.go
+++ b/pkg/proxyserver/controllers/clustermanagementaddon_controller.go
@@ -130,10 +130,6 @@ func (c *ClusterManagementAddonReconciler) Reconcile(ctx context.Context, reques
 	// get the related proxy configuration
 	config := &proxyv1alpha1.ManagedProxyConfiguration{}
 	if err := c.Client.Get(ctx, types.NamespacedName{Name: managedProxyConfigurationName}, config); err != nil {
-		if apierrors.IsNotFound(err) {
-			log.Info("Cannot find proxy-configuration", "name", managedProxyConfigurationName)
-			return reconcile.Result{}, nil
-		}
 		return reconcile.Result{}, err
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/proxyserver/controllers/clustermanagementaddon_controller.go
```diff
@@ -130,10 +130,7 @@ func (c *ClusterManagementAddonReconciler) Reconcile(ctx context.Context, reques
 	// get the related proxy configuration
 	config := &proxyv1alpha1.ManagedProxyConfiguration{}
 	if err := c.Client.Get(ctx, types.NamespacedName{Name: managedProxyConfigurationName}, config); err != nil {
-		if apierrors.IsNotFound(err) {
-			log.Info("Cannot find proxy-configuration", "name", managedProxyConfigurationName)
-			return reconcile.Result{}, nil
-		}
+		// if the config is not found, we should not return error, because the config might be created later or kube-apiserver may be not accessible at this moment.
```
skeeey, line 133: should return error to reconcile?

pkg/proxyserver/operator/eventhandler/proxyconfiguration_event.go
```diff
@@ -50,12 +50,11 @@ func findRelatedAddon(c client.Client, cfgName string, limitingInterface workque
 	}
 	for _, addon := range list.Items {
 		if addon.Name == common.AddonName {
-			if addon.Spec.AddOnConfiguration.CRName == cfgName {
```
xuezhaojun, line 53: AddonConfiguration is not used anymore.

pkg/proxyserver/controllers/clustermanagementaddon_controller.go
```diff
@@ -130,10 +130,6 @@ func (c *ClusterManagementAddonReconciler) Reconcile(ctx context.Context, reques
 	// get the related proxy configuration
 	config := &proxyv1alpha1.ManagedProxyConfiguration{}
 	if err := c.Client.Get(ctx, types.NamespacedName{Name: managedProxyConfigurationName}, config); err != nil {
-		if apierrors.IsNotFound(err) {
```
xuezhaojun, line 133: if the managedproxyconfig is not found, we should return error, because the config might be created later or kube-apiserver may be not accessible at this moment.

PR #158, opened 2023-04-19
workload partitioning for proxy-agent in managed cluster

pixelsoccupied:
similar to [service-proxy](https://github.com/stolostron/cluster-proxy-addon/pull/98), let the pod to pin to reserved cores. 

JIRA: OCPBUGS-11604

/cc @xuezhaojun


pixelsoccupied:
Using https://github.com/stolostron/cluster-proxy/pull/53 instead!

PR #167, opened 2023-09-05
Upgrade go.mod open-cluster-management.io api.

xuezhaojun:
https://github.com/open-cluster-management-io/ocm/issues/67


haoqing0110:
seems https://github.com/open-cluster-management-io/cluster-proxy/blob/main/hack/crd/cluster/managedclusterset.yaml
https://github.com/open-cluster-management-io/cluster-proxy/blob/main/hack/crd/bases/cluster.open-cluster-management.io_managedclustersets.yaml  are still using old version crd.

PR #171, opened 2023-12-06
update deployment securityContext

zhiweiyin318:


pkg/proxyserver/controllers/manifests.go
```diff
@@ -220,3 +229,13 @@ func newProxyServerRoleBinding(config *proxyv1alpha1.ManagedProxyConfiguration)
 	}
 
 }
+
+func truePtr() *bool {
```
qiujian16, line 233: use pointer library?

PR #176, opened 2024-03-04
Use Go 1.22

tamalsaha:


xuezhaojun:
Hi, @tamalsaha Thanks for the submit, since we have discovered same inconsistency in go1.22 https://github.com/open-cluster-management-io/managed-serviceaccount/pull/96#issuecomment-1976620241 , would you mind change to upgrade-to version of this PR to 1.21?

xuezhaojun:
The go.mod https://github.com/open-cluster-management-io/cluster-proxy/blob/ae20540551aaefc9a0f894795bd688ac6f5727aa/go.mod#L3 may also need to bump up.

PR #179, opened 2024-03-07
Update workflows.(change the github action used for release)

xuezhaojun:
Refer: the secrets.GITHUB_TOKEN https://docs.github.com/en/actions/security-guides/automatic-token-authentication

### DIFF
--- a/.github/workflows/go-release.yml
+++ b/.github/workflows/go-release.yml
@@ -11,20 +11,15 @@ env:
   GITHUB_REF: ${{ github.ref }}
   CHART_NAME: 'cluster-proxy'
 
-defaults:
-  run:
-    working-directory: go/src/open-cluster-management.io/cluster-proxy
-
 jobs:
   env:
     name: prepare release env
     runs-on: ubuntu-latest
     steps:
       - name: checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/go-release.yml
```diff
@@ -112,12 +112,20 @@ jobs:
...
+        uses: softprops/action-gh-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
-          tag: ${{ env.RELEASE_VERSION }}
-          artifacts: "go/src/open-cluster-management.io/cluster-proxy/release/*.tgz"
-          token: ${{ secrets.GITHUB_TOKEN }}
+          body_path: /home/runner/work/changelog.txt
+          files: |
+            release/*.tgz
```
zhujian7, line 125:
```suggestion
            go/src/open-cluster-management.io/cluster-proxy/release/*.tgz
```

since the default working-directory is go/src/open-cluster-management.io/cluster-proxy, https://github.com/open-cluster-management-io/cluster-proxy/pull/179/files#diff-29d508c8d0fc8d2fc3ea8e21d26ed76072f2de9d4a031e3307d41711e219bb76R16

xuezhaojun: Good point, I removed the default path to align with other repo and be simpler

PR #183, opened 2024-03-13
Add the hash value of signer into subject organizationunits

xuezhaojun:
Relate PR: https://github.com/open-cluster-management-io/ocm/pull/377

### DIFF
--- a/pkg/proxyagent/agent/agent.go
+++ b/pkg/proxyagent/agent/agent.go
@@ -2,6 +2,7 @@ package agent
 
 import (
 	"context"
+	"crypto/sha256"
 	"embed"
 	"encoding/base64"
 	"fmt"
@@ -73,6 +74,8 @@ func NewAgentAddon(
 		},
 	}
 	// Register the custom signer CSR option if V1 csr is supported
+	// caculate a hash value of signer ca data and add it to the organizationUnits of the subject
+	signerHash := sha256.Sum256(signer.CAData())
 	if v1CSRSupported {
 		regConfigs = append(regConfigs, addonv1alpha1.RegistrationConfig{
 			SignerName: ProxyAgentSignerName,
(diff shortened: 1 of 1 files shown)


pkg/proxyagent/agent/agent.go
```diff
@@ -73,6 +74,8 @@ func NewAgentAddon(
 		},
 	}
 	// Register the custom signer CSR option if V1 csr is supported
+	// caculate a hash value of signer ca data and add it to the organizationUnits of the subject
+	signerHash := sha256.Sum256(signer.CAData())
```
qiujian16, line 78: can we just use service endpoint?

xuezhaojun: service namespace and name is configurable in cluster-proxy: https://github.com/xuezhaojun/cluster-proxy/blob/8c762cca2bbc65e4fa5ecce0d5e8ec264fe4e3bd/pkg/proxyserver/controllers/manifests.go#L56-L57

qiujian16: i mean the endpoint exposed publicly, not internally

xuezhaojun: like route url in ocp case?

qiujian16: I think we need that as CN to sign the serving CA, isn't it?

xuezhaojun: I think you mean https://github.com/stolostron/cluster-proxy/blob/d60d4b3328e44114c283da343a3ef2638c2e0eed/charts/cluster-proxy/templates/managedproxyconfiguration.yaml#L15-L16

this entrypoint has multiple types, is is used as SANs when sign: https://github.com/stolostron/cluster-proxy/blob/d60d4b3328e44114c283da343a3ef2638c2e0eed/pkg/proxyserver/controllers/managedproxyconfiguration_controller.go#L388

But since it is in managedproxyconfiguration, there is chance the entrypoint change to another type.

qiujian16: If it is changed, do we need to re-sign the serving cert, and re-trigger the CSR?

xuezhaojun: No need to re-trigger CSR, the agent deployment will be updated to access the new endpoint but it can use the previous signed client cert.

The server certificates will be rotate by the managedproxyconfiguraion_controller on the hub side.

PR #205, opened 2024-08-19
Update golang and deps

qiujian16:


xuezhaojun:
@qiujian16  All tests passed, I will force merge it because the known reason that go 1.22 will count pkgs without _test.go as 0%.

xuezhaojun:
/lgtm

Suggest to force merge.

PR #214, opened 2024-11-03
Set failureThreshold to 3 for improved stability

RokibulHasan7:
addon-agent restarts due to a connection error: `"cannot connect once" err="rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial tcp 127.0.0.1:8091: connect: connection refused\""`


xuezhaojun:
@RokibulHasan7  The error log shows the connect error to the proxy-server,  change `failureThreshould` doesn't help in this case.

xuezhaojun:
@qiujian16 @RokibulHasan7 Suggest revert by: https://github.com/open-cluster-management-io/cluster-proxy/pull/216

xuezhaojun:
After discussion, we found it's not a bug, the root cause is the default entrypoint type is `PortForward` which is not suite to `k3s in a VM` case.

PR #286, opened 2026-05-08
Fix cluster-proxy service-proxy failback to hub authentication if managed cluster authentication fails

kurwang:
## Summary

- **Bug**: The service-proxy `processAuthentication` function returns immediately when the managed cluster `TokenReview` returns an error, never attempting hub authentication. This causes all hub-only users to receive "Unauthorized" when accessing managed cluster resources through the proxy.
- **Root cause**: Some Kubernetes distributions populate `TokenReview.Status.Error` for unrecognized tokens (unlike others which return `Authenticated: false` with no error). The `tokenReviewAuthenticator` surfaces this as a Go error, and `processAuthentication` treats any error from managed cluster auth as fatal — exiting before trying hub auth.
- **Fix**: Introduce an `ErrTokenNotAuthenticated` sentinel error. When a `TokenReview` explicitly rejects a token (`Status.Error` is non-empty), the error is wrapped with this sentinel. In `processAuthentication`, `errors.Is(err, ErrTokenNotAuthenticated)` distinguishes token rejection from infrastructure errors. Token rejections fall through to hub authentication; infrastructure errors remain fatal.

## Error Handling

| Scenario | Code Location | `err` value | `errors.Is(err, ErrTokenNotAuthenticated)` | Result |
|---|---|---|---|---|
| API call failed (network, TLS, etc.) | `token_authenticator.go:38-39` | plain error from `client.Create()` | `false` | Fatal, no fallback |
| Token rejected with `Status.Error` | `token_authenticator.go:49-51` | wraps `ErrTokenNotAuthenticated` sentinel | `true` | Fall through to hub |
| Token not recognized, no `Status.Error` | `token_authenticator.go:53-54` | `nil` | N/A | Fall through to hub |

The sentinel check is used in both auth paths: managed cluster (`service_proxy.go:369`) and hub (`service_proxy.go:391`).

## Impact

Any deployment where the managed cluster's Kubernetes distribution sets `TokenReview.Status.Error` for invalid tokens. Affects proxied requests (console access, pod logs, etc.) for users authenticated only on the hub.

## Changes

- `pkg/serviceproxy/token_authenticator.go`: Define `ErrTokenNotAuthenticated` sentinel error. Wrap it via `fmt.Errorf("%w")` when `TokenReview.Status.Error` is non-empty. Infrastructure errors from `client.Create()` are returned as-is.
- `pkg/serviceproxy/service_proxy.go`: Use `errors.Is(err, ErrTokenNotAuthenticated)` to distinguish token rejection (fall through to hub auth) from infrastructure errors (fatal). Applied consistently to both managed cluster and hub authentication paths.
- `pkg/serviceproxy/token_authenticator_test.go`: Updated and added tests covering:
  - `TestTokenReviewAuthenticator_StatusError` — verifies `Status.Error` produces an error wrapping `ErrTokenNotAuthenticated`
  - `TestProcessAuthentication_ManagedClusterFatalError` — infrastructure error is fatal, hub auth not attempted
  - `TestProcessAuthentication_OpenShiftTokenReviewError_FallsBackToHub` — token rejection falls through to hub auth with correct impersonation
  - `TestProcessAuthentication_HubAuthError` — hub infrastructure error is fatal with error preserved

## Test plan

- [x] Unit tests pass (`go test ./pkg/serviceproxy/`)
- [x] Verified on live environment: deployed to managed cluster, confirmed hub user can access resources through cluster-proxy (previously returned 401 Unauthorized)
- [x] Verify no regression for managed-cluster-native tokens (tokens recognized by the managed cluster should still work without hitting hub auth)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authentication flow so certain managed-cluster authentication rejections are treated as non-fatal and correctly fall back to hub token review; other hub failures remain fatal.
  * Clarified error handling to preserve original failure context when authentication is fatal.

* **Tests**
  * Expanded tests to cover known rejection patterns, unknown error cases, fatal vs non-fatal classification, fallback behavior, and correct impersonation handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/open-cluster-management-io/cluster-proxy/pull/286?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/pkg/serviceproxy/service_proxy.go
+++ b/pkg/serviceproxy/service_proxy.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
+	"errors"
 	"fmt"
 	"net"
 	"net/http"
@@ -365,8 +366,12 @@ func (s *serviceProxy) processAuthentication(ctx context.Context, req *http.Requ
 	// try managed cluster authentication first
 	managedClusterResp, managedClusterAuthenticated, err := s.managedClusterAuthenticator.AuthenticateToken(ctx, token)
 	if err != nil {
-		logger.Error(err, "managed cluster authentication failed")
-		return fmt.Errorf("managed cluster authentication failed: %v", err)
+		if errors.Is(err, ErrTokenNotAuthenticated) {
+			logger.V(4).Info("managed cluster token not authenticated, trying hub", "error", err)
+			managedClusterAuthenticated = false
+		} else {
+			return fmt.Errorf("managed cluster authentication failed: %v", err)
+		}
 	}
 
 	if managedClusterAuthenticated {
@@ -383,8 +388,13 @@ func (s *serviceProxy) processAuthentication(ctx context.Context, req *http.Requ
 		// try hub authentication
 		hubResp, hubAuthenticated, err := s.hubAuthenticator.AuthenticateToken(ctx, token)
 		if err != nil {
-			logger.Error(err, "hub cluster authentication failed")
-			return fmt.Errorf("authentication failed: managed cluster auth: not authenticated, hub cluster auth error: %v", err)
+			if errors.Is(err, ErrTokenNotAuthenticated) {
+				logger.V(4).Info("hub cluster token not authenticated", "error", err)
+				hubAuthenticated = false
+			} else {
+				logger.Error(err, "hub cluster authentication failed")
+				return fmt.Errorf("authentication failed: managed cluster auth: not authenticated, hub cluster auth error: %v", err)
+			}
 		}
 		logger.V(4).Info("hub cluster authentication result",
 			"authenticated", hubAuthenticated,
--- a/pkg/serviceproxy/token_authenticator.go
+++ b/pkg/serviceproxy/token_authenticator.go
@@ -2,6 +2,7 @@ package serviceproxy
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	authenticationv1 "k8s.io/api/authentication/v1"
@@ -11,6 +12,11 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
 )
+
+// ErrTokenNotAuthenticated is returned when a TokenReview explicitly rejects a
+// token (Status.Error is non-empty). Callers use errors.Is to distinguish this
+// from infrastructure errors (network, TLS, etc.).
+var ErrTokenNotAuthenticated = errors.New("token not authenticated")
 
 // tokenReviewAuthenticator implements authenticator.Token by calling the
 // Kubernetes TokenReview API against a specific cluster.
@@ -40,10 +46,11 @@ func (a *tokenReviewAuthenticator) AuthenticateToken(ctx context.Context, token 
 		"groups", tokenReview.Status.User.Groups,
 	)
 
+	if tokenReview.Status.Error != "" {
+		return nil, false, fmt.Errorf("%s TokenReview: %s: %w", a.name, tokenReview.Status.Error, ErrTokenNotAuthenticated)
+	}
+
 	if !tokenReview.Status.Authenticated {
-		if tokenReview.Status.Error != "" {
-			return nil, false, fmt.Errorf("%s TokenReview error: %s", a.name, tokenReview.Status.Error)
-		}
 		return nil, false, nil
 	}
 
--- a/pkg/serviceproxy/token_authenticator.go
+++ b/pkg/serviceproxy/token_authenticator.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
 	authenticationv1 "k8s.io/api/authentication/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -14,9 +15,29 @@ import (
 )
 
 // ErrTokenNotAuthenticated is returned when a TokenReview explicitly rejects a
-// token (Status.Error is non-empty). Callers use errors.Is to distinguish this
-// from infrastructure errors (network, TLS, etc.).
+// token with a known authentication error. Callers use errors.Is to distinguish
+// this from infrastructure errors (network, TLS, etc.).
 var ErrTokenNotAuthenticated = errors.New("token not authenticated")
+
+// authRejectionPatterns contains substrings of TokenReview Status.Error that
+// indicate the token was rejected by the authenticator. Observed from:
+//   - Kubernetes v1.35: "invalid bearer token"
+//   - OpenShift 4.x:    "[invalid bearer token, token lookup failed]"
+var authRejectionPatterns = []string{
+	"invalid bearer token",
+}
+
+// isAuthRejection reports whether a TokenReview Status.Error indicates an
+// authentication rejection rather than an infrastructure failure.
+func isAuthRejection(statusError string) bool {
+	lower := strings.ToLower(statusError)
+	for _, pattern := range authRejectionPatterns {
+		if strings.Contains(lower, pattern) {
+			return true
+		}
+	}
+	return false
+}
 
 // tokenReviewAuthenticator implements authenticator.Token by calling the
 // Kubernetes TokenReview API against a specific cluster.
@@ -47,7 +68,10 @@ func (a *tokenReviewAuthenticator) AuthenticateToken(ctx context.Context, token 
 	)
 
 	if tokenReview.Status.Error != "" {
-		return nil, false, fmt.Errorf("%s TokenReview: %s: %w", a.name, tokenReview.Status.Error, ErrTokenNotAuthenticated)
+		if isAuthRejection(tokenReview.Status.Error) {
+			return nil, false, fmt.Errorf("%s TokenReview: %s: %w", a.name, tokenReview.Status.Error, ErrTokenNotAuthenticated)
+		}
+		return nil, false, fmt.Errorf("%s TokenReview: %s", a.name, tokenReview.Status.Error)
 	}
 
 	if !tokenReview.Status.Authenticated {
--- a/pkg/serviceproxy/token_authenticator_test.go
+++ b/pkg/serviceproxy/token_authenticator_test.go
@@ -2,6 +2,7 @@ package serviceproxy
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"strings"
@@ -72,7 +73,7 @@ func TestTokenReviewAuthenticator_Unauthenticated(t *testing.T) {
 
 func TestProcessAuthentication_ManagedClusterToken(t *testing.T) {
 	s := &serviceProxy{
-		enableImpersonation:         true,
+		enableImpersonation: true,
 		managedClusterAuthenticator: authenticator.TokenFunc(func(ctx context.Context, token string) (*authenticator.Response, bool, error) {
 			return &authenticator.Response{User: &user.DefaultInfo{Name: "mc-user"}}, true, nil
 		}),
@@ -303,6 +304,9 @@ func TestTokenReviewAuthenticator_StatusError(t *testing.T) {
 	if resp != nil {
 		t.Fatal("expected nil response")
 	}
+	if !errors.Is(err, ErrTokenNotAuthenticated) {
+		t.Fatalf("expected ErrTokenNotAuthenticated, got: %v", err)
+	}
 	if !strings.Contains(err.Error(), "Credentials are expired") {
 		t.Fatalf("expected Status.Error in error message, got: %v", err)
 	}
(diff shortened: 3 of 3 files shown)


pkg/serviceproxy/service_proxy.go
mikeshng: We should take a look here as well right to be consistent?

CC @tesshuflower @rokej

pkg/serviceproxy/service_proxy.go
```diff
@@ -365,8 +365,9 @@ func (s *serviceProxy) processAuthentication(ctx context.Context, req *http.Requ
 	// try managed cluster authentication first
 	managedClusterResp, managedClusterAuthenticated, err := s.managedClusterAuthenticator.AuthenticateToken(ctx, token)
 	if err != nil {
-		logger.Error(err, "managed cluster authentication failed")
-		return fmt.Errorf("managed cluster authentication failed: %v", err)
+		// Treat managed cluster TokenReview errors as "not authenticated" rather than fatal.
```
mikeshng: Ideally AuthenticateToken should return an error that is specific to NotAuthenticated then if the error is NotAuthenticated, we mark it as false, otherwise it shoul dstill be fatal error.

pkg/serviceproxy/service_proxy.go
rokej: You are right @mikeshng. This code block should log the error, set hubAuthenticated=false and let the `if !hubAuthenticated` line return the error or remove that block.

pkg/serviceproxy/service_proxy.go
kurwang: makes sense will do

pkg/serviceproxy/service_proxy.go
kurwang: ack

pkg/serviceproxy/service_proxy.go
```diff
@@ -365,8 +366,12 @@ func (s *serviceProxy) processAuthentication(ctx context.Context, req *http.Requ
 	// try managed cluster authentication first
 	managedClusterResp, managedClusterAuthenticated, err := s.managedClusterAuthenticator.AuthenticateToken(ctx, token)
 	if err != nil {
-		logger.Error(err, "managed cluster authentication failed")
-		return fmt.Errorf("managed cluster authentication failed: %v", err)
+		if apierrors.IsUnauthorized(err) {
```
kurwang: im hoping i understood what you meant here @mikeshng 
if there is an err and is an authentication error it can continue but if there is another error that should be fatal

mikeshng:
@kurwang yea my comment is a generic golang err handling feedback. You will have to double check the code (beyond the diffs here) and even test your code changes to make sure it all works well. From me glancing at the recent changes, I am not sure if unauthorized can be use as unauthenticated, they might mean different things (ie. logged in but not enough permission to do stuffs vs not logged in at all so can't do anything period.) 

Example of the generic golang err handling:
```
if err != nil
   if the err is something I am expecting deal with it accordingly.
   else the err is not something I am expecting (could be literally anything) then bubble the error up
```
You will have to check yourself to see if this can even be accomplish in this case. Maybe it's not possible and your initial changes is the only way. 

I am going to leave this PR review to the good hands of @rokej and @tesshuflower

kurwang:
@mikeshng sounds good ill look closer to this detail and report results thanks

kurwang:
New commit with the following changes @rokej @tesshuflower let me know what you think of this approach similar to my first one

based on Mikes previous comments I investigated a bit more on TokenReivew
https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-review-v1/

which seems to me that it focuses on authentication and Not anything with authorization
so I am going to follow the same err handling pattern but change the error type itself so it deals with authentication and aligns with TokenReview

This approach helps distinguish other errors vs authentication error based on the tokenreviewspec

## Error Handling Summary

| Scenario | Code Location | `err` value | `errors.Is(err, ErrTokenNotAuthenticated)` | Result |
|---|---|---|---|---|
| API call failed (network, TLS, etc.) | `token_authenticator.go:38-39` | plain error from `client.Create()` | `false` | Fatal, no fallback |
| Token rejected with `Status.Error` (OpenShift) | `token_authenticator.go:50-51` | wraps `ErrTokenNotAuthenticated` sentinel | `true` | Fall through to hub |
| Token not recognized, no `Status.Error` (standard K8s) | `token_authenticator.go:53` | `nil` | N/A | Fall through to hub |

kurwang:
env test

## Live Environment Test Results

### Test 1: Hub user (`kube:admin`) accessing managed cluster through proxy

This tests the hub authentication fallback — the managed cluster rejects the hub token, and the fix falls through to hub auth.

```bash
oc exec -n multicluster-engine curl-test -- curl -sk \
  -H "Authorization: Bearer <hub-kube-admin-token>" \
  "https://cluster-proxy-addon-user.multicluster-engine.svc:9092/sno-2-xsr74/api/v1/namespaces?limit=1"
```

**Result: 200 OK** — hub fallback worked, namespaces returned.

```json
{
  "kind": "NamespaceList",
  "apiVersion": "v1",
  "metadata": {
    "resourceVersion": "4809331",
    "remainingItemCount": 81
  },
  "items": [
    {
      "metadata": {
        "name": "default",
        "uid": "80ef430d-6553-4a45-951a-57029ee5c42a"
      },
      "status": { "phase": "Active" }
    }
  ]
}
```

### Test 2: Managed cluster user (`kurwang`) accessing managed cluster through proxy

This tests that managed cluster native tokens still authenticate directly without hub fallback (no regression).

```bash
oc exec -n multicluster-engine curl-test -- curl -sk \
  -H "Authorization: Bearer <managed-cluster-user-token>" \
  "https://cluster-proxy-addon-user.multicluster-engine.svc:9092/sno-2-xsr74/api/v1/namespaces?limit=1"
```

**Result: 403 Forbidden** — token was authenticated as `kurwang`, but user lacks RBAC permission to list namespaces.

```json
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {},
  "status": "Failure",
  "message": "namespaces is forbidden: User \"kurwang\" cannot list resource \"namespaces\" in API group \"\" at the cluster scope",
  "reason": "Forbidden",
  "details": { "kind": "namespaces" },
  "code": 403
}
```

### Test 3: Managed cluster admin (`kube:admin`) accessing managed cluster through proxy

This tests that managed cluster admin tokens authenticate directly and have full access (no regression).

```bash
oc exec -n multicluster-engine curl-test -- curl -sk \
  -H "Authorization: Bearer <managed-cluster-admin-token>" \
  "https://cluster-proxy-addon-user.multicluster-engine.svc:9092/sno-2-xsr74/api/v1/namespaces?limit=1"
```

**Result: 200 OK** — token authenticated directly on managed cluster, namespaces returned.

### Summary

| Test | Token | Auth path | Result |
|---|---|---|---|
| Hub `kube:admin` | Hub token | Managed cluster rejects → fallback to hub → impersonation | 200 OK |
| Managed cluster `kurwang` | sno-2 token | Managed cluster auth succeeds directly | 403 Forbidden (no RBAC) |
| Managed cluster `kube:admin` | sno-2 admin token | Managed cluster auth succeeds directly | 200 OK |

kurwang:
@rokej @tesshuflower whenever you get a chance can you review my approach for cluster proxy checking authentication for managedcluster then hub?

rokej:
@kurwang Great! It looks good to me. Waiting on @tesshuflower 's review as well.

tesshuflower:
/lgtm

👍

kurwang:
@rokej does this need your approval to start the github workflows?

pkg/serviceproxy/token_authenticator.go
```diff
@@ -42,7 +48,7 @@ func (a *tokenReviewAuthenticator) AuthenticateToken(ctx context.Context, token
 
 	if !tokenReview.Status.Authenticated {
 		if tokenReview.Status.Error != "" {
-			return nil, false, fmt.Errorf("%s TokenReview error: %s", a.name, tokenReview.Status.Error)
+			return nil, false, fmt.Errorf("%s TokenReview: %s: %w", a.name, tokenReview.Status.Error, ErrTokenNotAuthenticated)
```
qiujian16, line 72: is there a case that tokenReview.Status.Authenticated=true and err != nil? I do not see this as a possible case.

And how do you diff the error, I think if it returns error, tokenReview.Status.Authenticated is always false. So the caller always call hub when err is not nil.

kurwang: right for this case it isnt possible I believe line 38 addresses this part
> is there a case that tokenReview.Status.Authenticated=true and err != nil? I do not see this as a possible case.

for this are you saying that regardless of the error, tokenReview.Status.Authenticated is always false
so the caller should always fallback to hub?
> And how do you diff the error, I think if it returns error, tokenReview.Status.Authenticated is always false. So the caller always call hub when err is not nil.

if so i agree this can be simplified but my reasoning for this distinction on error was because it address any empty status but still calls hub either way

tesshuflower:
The latest update looks good to me, checking `xtokenReview.Status.Error` first and then if no error, then checking for `!tokenReview.Status.Authenticated`.

so I'm good with this, but will defer to @qiujian16 for the final review

pkg/serviceproxy/token_authenticator.go
qiujian16: my point is when err != nil, tokenReview.Status.Authenticated is always false. So line 49 is redundant, however, the question is we need to differentiate error types. For instance, if apiserver of spoke is done, it also returns an error, and goes to hub authentication process which seems unreasonable.

kurwang: sorry i might be a bit confused so just tracking my steps and error handling here

so I do understand that line 49 is redundant i think the updated code has a better handling of errors

if there is an api server from spoke for example I believe that is handled in the earlier line of code at line 38-39. Hub auth is never attempted here when processAuthentication calls it

Line 49-51 only runs when the API call **succeeds** but the token is explicitly rejected (`Status.Error` is set). That error wraps `ErrTokenNotAuthenticated`, so the caller falls through to hub auth.

So the two error paths are:
- **Line 38-39**: infrastructure failure (spoke down) → plain error → fatal, no hub fallback
- **Line 49-51**: token rejection (`Status.Error` set) → wraps sentinel → falls through to hub

Then hub performs similar error checks on if it is authenticated or not

qiujian16: it is not a good example. But my point is authenticated==false does not always equal to authenticated==false && err != "". E.g. if there is a webhook authenticator and the webhook is not working, will it return both authenticated==false  and err != "", should the proxy go to hub side authentication in this case? I think not.

kurwang: gotcha ok i think i see what you mean thanks

hopefully now i understand your comments and it makes sense to me
so when tokenReview.Status.Error != nil there is a possibility of what kind of error it is

for example it is a webhook problem so authenticated==false and err != "" and not always should go to hub side authentication

but it can also just be an invalid token where it is still authenticated==false and err != "" and should go to hub side authentication correcct?

if so do you have any suggestions on how to handle this case?
im not as familiar with the api but from my understanding the tokenReview.Status.Error can have different outputs so handling this is a bit tricky handling any kind of error

kurwang: rereading and maybe im interpreting your comments wrong 
i added a different code change in the pr can you see if this is what you meant also?

if err != "" regardless of authneticated == false (what you mentioned) from token creation on line 38

proxy would not go to hub authentication

```go
	if err != nil {
		return nil, false, err
	}
```

but again maybe this isnt what you are talking about

pkg/serviceproxy/token_authenticator.go
```diff
@@ -40,10 +46,11 @@ func (a *tokenReviewAuthenticator) AuthenticateToken(ctx context.Context, token
 		"groups", tokenReview.Status.User.Groups,
 	)
 
+	if tokenReview.Status.Error != "" {
+		return nil, false, fmt.Errorf("%s TokenReview: %s: %w", a.name, tokenReview.Status.Error, ErrTokenNotAuthenticated)
```
qiujian16: I think you need to parse the error message and return ErrTokenNotAuthenticated for only certain conditions.

kurwang: im not that familiar with what Error message can be or atleast the important ones and i cant find it in the docs
are there important ones that should be captured beside token ones?

from my testing I see that tokenReview.Status.Error != "" can capture

invalid token outputs
`"status": {
        "error": "[invalid bearer token, token lookup failed]",
        "user": {}
    }`
    
expired token generates
`{
  "status": {
    "error": "[invalid bearer token, service account token has expired, token lookup failed]",
    "user": {}
  }
}`

it seems like from my testing fatal errors from api gets caught during the token create in line 38

pkg/serviceproxy/token_authenticator.go
```diff
@@ -12,6 +14,31 @@ import (
...
+// indicate the token was rejected by the authenticator. Observed from:
+//   - Kubernetes v1.35: "invalid bearer token"
+//   - OpenShift 4.x:    "[invalid bearer token, token lookup failed]"
+var authRejectionPatterns = []string{
+	"invalid bearer token",
+}
+
+// isAuthRejection reports whether a TokenReview Status.Error indicates an
+// authentication rejection rather than an infrastructure failure.
+func isAuthRejection(statusError string) bool {
```
kurwang, line 32: @qiujian16 Hi thanks for the pr reviews would you mind taking another look at this
 im taking what you mentioned and went with this approach where im parsing tokenReview.Status.Error and if it isnt a token issue it will return fatal

The token related errors I found were ones that were invalid and expired which looked like
invalid token outputs
`"status": { "error": "[invalid bearer token, token lookup failed]", "user": {} }`

expired token generates
`{ "status": { "error": "[invalid bearer token, service account token has expired, token lookup failed]", "user": {} } }`

qiujian16 (commented):
/approve
/lgtm

mikeshng (commented):
@kurwang did you add the latest commit to retrigger the CI? I don't see any additional changes. I will lgtm if that's the case. Thanks!

kurwang:
@mikeshng Hi yes i did just to retrigger the CI there wasnt any additional changes 
the e2e test was failing and didnt seem to be related to my code changes

mikeshng (approved):
/lgtm

@kurwang ok please don't commit for now and I will keep retrying the e2e and see how that goes. Thanks!

mikeshng:
@tesshuflower the e2e is failing. Could you please look into this? Thank you!

PR #297, opened 2026-06-16
Stabilize ConfigMap watch e2e test

kahirokunn:
## Summary
Stabilize the ConfigMap watch e2e test by stopping the watcher, waiting for the `cluster-proxy-test` `Added` event with `Eventually`, and deleting the test ConfigMap after it is observed.

## Related issue(s)
None.

### DIFF
--- a/test/e2e/connect_test.go
+++ b/test/e2e/connect_test.go
@@ -17,9 +17,9 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 
 	corev1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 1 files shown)


kahirokunn:

test/e2e/connect_test.go
```diff
@@ -177,26 +177,24 @@ var _ = Describe("Requests through Cluster-Proxy", Label("serviceproxy", "connec
...
-				Expect(err).To(BeNil())
-			default:
-				Fail("Failed to received a pod create event")
-			}
+			Eventually(watcher.ResultChan(), timeout, time.Second).Should(Receive(Satisfy(func(event watch.Event) bool {
+				configMap, ok := event.Object.(*corev1.ConfigMap)
+				return ok && event.Type == watch.Added && configMap.Name == "cluster-proxy-test"
+			})))
+
+			err = hubKubeClient.CoreV1().ConfigMaps(hubInstallNamespace).Delete(context.Background(), "cluster-proxy-test", v1.DeleteOptions{})
```
tesshuflower, line 196: Not a big deal (and not caused by your change), but wonder if we should make sure to always cleanup (in a defer or something of the sort) even if the test above in the eventually fails.

I'm good with the change though and this is definitely an improvement

PR #299, opened 2026-06-16
:sparkles: Validate ManagedProxyConfiguration toleration operators

kahirokunn:
## Summary
- Add CRD validation for ManagedProxyConfiguration toleration operators while preserving empty-operator compatibility.

## Related issue(s)
None.


kahirokunn:

mikeshng (commented):
I fully understand the motivation behind this PR but it does introduce backward compatibility risk (restricting values) so let's get at least 2 maintainers to approve.

/hold

/assign @qiujian16

kahirokunn:
Thanks for your review 🙏 Okay 💪

kahirokunn:
just fixed conflict

kahirokunn:
@qiujian16 Hi ✋ PTAL. Thank you 🙏

pkg/apis/proxy/v1alpha1/managedproxyconfiguration_types.go
```diff
@@ -169,7 +169,54 @@ type NodePlacement struct {
 	// the triple <key,value,effect> using the matching operator <operator>.
 	// The default is an empty list.
 	// +optional
-	Tolerations []v1.Toleration `json:"tolerations,omitempty"`
```
mikeshng, line 172: Coming back to this PR. I actually like the fact that we are importing from v1 core API instead of defining our own. So we gain validation but lose out on the convenience (and maybe some advantages as well) of directly importing from v1 API. 

/assign @tesshuflower WDYT?

tesshuflower: I think I agree with @mikeshng here, I'd prefer not to add our own custom tolerations if possible.

kahirokunn: Regarding this point, there are pros and cons, and I've been unsure myself, so I'll close this case.

PR #302, opened 2026-06-16
:bug: Stabilize e2e readiness checks

kahirokunn:
## Summary

- Wait for the proxy-agent Deployment to finish rolling out before checking e2e readiness.
- Check only pods from the latest addon configuration and skip pods that are already being deleted.
- Save the existing cluster-proxy addon config before the deployment config e2e test, then restore it afterward so later checks start from the original state.
- Include `stderr` from failed `exec` calls in connection test errors so failures are easier to debug.

### DIFF
--- a/test/e2e/connect_test.go
+++ b/test/e2e/connect_test.go
@@ -221,8 +221,8 @@ var _ = Describe("Requests through Cluster-Proxy", Label("serviceproxy", "connec
 				Stderr: &stderr,
 				Tty:    false,
 			})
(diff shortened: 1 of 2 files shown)


mikeshng (commented):
It looks ok to me but might need a more careful look to ensure no downstream impact.

/assign @tesshuflower

kahirokunn:
Split this PR so the e2e readiness changes can be reviewed separately from the Dockerfile build cleanup.

- The Dockerfile-only change moved to #311.

PR #305, opened 2026-06-17
Harden Helm chart pod security

kahirokunn:
## Summary
- fix the embedded addon-agent chart so default Helm render no longer emits invalid YAML
- set pod-level `seccompProfile.type: RuntimeDefault` on cluster-proxy workloads and e2e pods
- add a Helm chart CI target and presubmit job to lint/render both charts
- update GitHub Actions references to latest releases and pin them by commit SHA
- add Dependabot version updates for SHA-pinned GitHub Actions

## Details
The addon-agent chart defaulted to an empty image tag for some containers, which rendered `image: ...:` and caused Helm parsing to fail. The chart now uses `latest` as the standalone fallback while runtime values from the top-level chart still provide release-tagged images.

The top-level chart already defaults image tags to `v<Chart.Version>`, and release workflows verify that the chart version matches the release tag without the `v` prefix. That keeps normal releases on `vX.Y.Z` image tags while leaving addon-agent standalone rendering usable.

The Helm presubmit pins setup-helm to a current action SHA and Helm v3.20.1 so CI does not accidentally install an old Helm prerelease. Other workflow actions are also SHA-pinned with version comments for traceability.

Dependabot is configured for weekly grouped GitHub Actions updates from `.github/workflows`, so future action update PRs can move the pinned SHAs while preserving version comments.

## CodeRabbit follow-up
Addressed CodeRabbit's presubmit workflow hardening feedback in a follow-up commit. The presubmit workflow runs on `pull_request` and executes repository Make targets, so it should use the minimum GitHub token permissions needed, avoid persisting checkout credentials in git config, and avoid `setup-go`'s default cache writes for PR-executed jobs.

- Added top-level `permissions: contents: read` and `persist-credentials: false` on all `actions/checkout` steps: https://github.com/open-cluster-management-io/cluster-proxy/pull/305#discussion_r3425728676
- Added `cache: false` on all `actions/setup-go` steps in `go-presubmit.yml`: https://github.com/open-cluster-management-io/cluster-proxy/pull/305#discussion_r3425728686

## Validation
- `go test ./pkg/proxyagent/agent ./pkg/proxyserver/controllers`
- `go test ./pkg/...`
- `make test-helm`
- workflow YAML parse check and SHA-pin check for every `uses:` entry
- `git diff --check`
- `yq '.' .github/workflows/go-presubmit.yml`
- custom workflow hardening check for permissions, checkout credential persistence, and setup-go cache settings

## Summary by CodeRabbit

* **Security**
  * Applied pod-level RuntimeDefault seccomp profile configuration to cluster proxy deployments (manager, user) and test pod manifests.

* **CI/CD & Automation**
  * Configured automated dependency updates for GitHub Actions repositories using Dependabot with weekly scheduling.
  * Pinned third-party action versions to specific commit hashes to ensure build consistency and reproducibility.
  * Added Helm chart linting and template validation as new build pipeline steps.

* **Tests**
  * Implemented validation tests to verify seccomp profile configuration in generated deployments.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,16 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+      time: "09:00"
+      timezone: "Asia/Tokyo"
+    groups:
+      github-actions:
+        patterns:
+          - "*"
+    commit-message:
+      prefix: "ci"
+      include: "scope"
--- a/.github/workflows/chart-upload.yml
+++ b/.github/workflows/chart-upload.yml
@@ -20,7 +20,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: checkout code
-        uses: actions/checkout@v4
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
         with:
           fetch-depth: 1
       - name: get release version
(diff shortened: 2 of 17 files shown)


kahirokunn:
The PR had gotten too big, so I took it apart.

PR #308, opened 2026-06-17
:seedling: pin GitHub Actions and add Dependabot updates

kahirokunn:
## Summary
- Pin existing GitHub Actions workflow `uses:` references to commit SHAs with version comments.
- Add Dependabot configuration for weekly grouped GitHub Actions updates.
- Harden the presubmit workflow based on CodeRabbit feedback from the split-out original PR.
- Remove the stale `third_party/apiserver-network-proxy` submodule reference.

## Split-out follow-up
- The clusteradm pinning and `hack/tools` module changes were moved to #312 for separate review and discussion.

## CodeRabbit follow-up
This PR carries the GitHub Actions hardening feedback that CodeRabbit originally reported on the PR. The presubmit workflow runs on `pull_request` and executes repository Make targets, so it should use the minimum GitHub token permissions needed, avoid persisting checkout credentials in git config, and avoid `setup-go`'s default cache writes for PR-executed jobs.

- Added top-level `permissions: contents: read` and `persist-credentials: false` on all `actions/checkout` steps: https://github.com/open-cluster-management-io/cluster-proxy/pull/305#discussion_r3425728676
- Added `cache: false` on all `actions/setup-go` steps in `go-presubmit.yml`: https://github.com/open-cluster-management-io/cluster-proxy/pull/305#discussion_r3425728686

## Validation
- `git range-diff origin/main..backup/ci/github-actions-pinning-before-drop-clusteradm-tools-20260618141541 origin/main..ci/github-actions-pinning`
- `git diff --check $(git merge-base origin/main HEAD)..HEAD`
- `yq '.' .github/workflows/go-presubmit.yml`

## Summary by CodeRabbit

* **Chores**
  * Strengthened continuous integration security and consistency by pinning all external GitHub Actions and dependencies to specific verified versions
  * Configured automated weekly updates for development tool dependencies
  * Updated internal project dependencies and third-party component references

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -33,3 +33,13 @@ updates:
     commit-message:
       prefix: "ci"
       include: "scope"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    commit-message:
+      prefix: ":seedling:"
+    groups:
+      github-actions:
+        patterns:
+          - "*"
--- a/.github/workflows/go-presubmit.yml
+++ b/.github/workflows/go-presubmit.yml
@@ -22,9 +22,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: checkout code
-        uses: actions/checkout@v4
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
       - name: install Go
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
         with:
           go-version: ${{ env.GO_VERSION }}
       - name: build
@@ -34,9 +34,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
       - name: install Go
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
         with:
           go-version: ${{ env.GO_VERSION }}
       - name: verify
@@ -60,15 +60,15 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: checkout code
-        uses: actions/checkout@v4
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
       - name: install Go
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
         with:
           go-version: ${{ env.GO_VERSION }}
       - name: unit
         run: make test
       - name: report coverage
-        uses: codecov/codecov-action@fb8b3582c8e4def4969c97caa2f19720cb33a72f
+        uses: codecov/codecov-action@fb8b3582c8e4def4969c97caa2f19720cb33a72f # v7.0.0
         with:
           token: ${{ secrets.CODECOV_UPLOAD_TOKEN }}
           files: ./cover.out
@@ -82,9 +82,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: checkout code
-        uses: actions/checkout@v4
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
       - name: install Go
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
         with:
           go-version: ${{ env.GO_VERSION }}
       - name: integration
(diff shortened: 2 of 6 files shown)


kahirokunn:

.github/workflows/go-presubmit.yml
```diff
@@ -81,15 +96,18 @@ jobs:
...
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
         with:
           go-version: ${{ env.GO_VERSION }}
+          cache: false
       - name: Install clusteradm
         run: curl -L https://raw.githubusercontent.com/open-cluster-management-io/clusteradm/main/install.sh | bash
       - name: Create k8s Kind Cluster
-        uses: helm/kind-action@v1.2.0
+        uses: helm/kind-action@ef37e7f390d99f746eb8b610417061a60e82a6cc # v1.14.0
```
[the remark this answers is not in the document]
kahirokunn: Thanks, addressed. I removed the mutable `main` installer and now build `clusteradm` from a pinned Go tools module under `hack/tools` (`open-cluster-management.io/clusteradm v1.3.1`). I also added `/hack/tools` to Dependabot's gomod config so future clusteradm updates are automated.

mikeshng (commented):
I am onboard with the pin GitHub actions and adding the dependency bots to help with automation. 

My concern is the clusteradm changes is impactful enough to be its own PR and we can discuss more there on it and let the Pining + DepBot merge first.

WDYT?

kahirokunn:
@mikeshng Thanks for your review :pray: 
clusteradm is also essentially just pinned without depending on main, but should I still split this into separate PRs?

When pinning, I want to use Dependabot for automatic updates. However, Dependabot does not automatically update URLs like `https://raw.githubusercontent.com/open-cluster-management-io/clusteradm/main/install.sh`, so I introduced hack/tools to pin the version while still allowing for Dependabot automation. I adopted this same approach for hack/tools as used in other k8s-sigs repositories (such as cluster-api).

mikeshng:
Yea, I agree with you if this is a convention that is done in other k8s repos. This might actually be a good practice to follow for all the other repos that is currently downloading the clusteradm binary. 

I will withdraw my previous comment and let other maintainers take a look.

/assign @tesshuflower @qiujian16

kahirokunn:
Here is an example.
https://github.com/kubernetes-sigs/cluster-api/tree/main/hack/tools

 I think this is a good pattern. As you mentioned, I agree that it would be fine to adopt it in other repositories as well.

tesshuflower:
Not trying to be difficult here, but I think I have to agree with @mikeshng 's original thought about possibly removing the clusteradm update from this?

The current way of getting clusteradm uses that install.sh script, which will get the `latest` - If we pin to a specific version, then we're not getting the latest version of all the other projects in the e2e, so this is more than pinning to our existing version, it changes the behavior.

Now, to have reproducible e2e tests, it probably is best to pin to a specific clusteradm version as you did? not sure the historic reason behind getting the latest always, but perhaps it's to get ahead of integration errors with updates to code in the other repos?

kahirokunn:
fixed conflict

tesshuflower:
As per my comment above, I think it would be best not to have the clusteradm pinning as part of this PR - we have to be careful about pinning the version as previously these repos always tested against the latest.

kahirokunn:
Okay. Separated 👍

.github/workflows/go-presubmit.yml
```diff
@@ -1,5 +1,8 @@
 name: Go
 
+permissions:
```
tesshuflower, line 3: should we also do this for dco.yml, go-postsubmit.yml?

.github/workflows/go-presubmit.yml
```diff
@@ -22,39 +25,48 @@ jobs:
...
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
         with:
           go-version: ${{ env.GO_VERSION }}
+          cache: false
       - name: unit
         run: make test
       - name: report coverage
-        uses: codecov/codecov-action@fb8b3582c8e4def4969c97caa2f19720cb33a72f
+        uses: codecov/codecov-action@fb8b3582c8e4def4969c97caa2f19720cb33a72f # v7.0.0
```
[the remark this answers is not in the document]
tesshuflower: This is weird - I think the sha in the PR is correct?

.github/dependabot.yml
```diff
@@ -0,0 +1,16 @@
+version: 2
```
tesshuflower, line 1: Do you think we should use the same settings for this and managed-serviceaccount?  Any reason to keep them different?

.github/workflows/go-presubmit.yml
kahirokunn: yes, fixed. thank you 🙏

.github/dependabot.yml
kahirokunn: Thank you. I fixed 🙏

kahirokunn:

tesshuflower:
/hold

Oops I just remembered, should we update the dependabot pr titles to use `:seedling:` here as well?

tesshuflower:
my hold was too late, but it's ok... - I was wrong anyway, it is already configured as such

PR #306, opened 2026-06-17
:seedling: Add Helm chart presubmit checks

kahirokunn:
## Summary

- Add a `make test-helm` target that lints and renders the cluster-proxy Helm charts.
- Exercise the default deployment path and service-proxy rendering for both the top-level `cluster-proxy` chart and the embedded `addon-agent` chart.
- Add a dedicated least-privilege `helm` presubmit job to run the chart checks.

## Summary by CodeRabbit

* **Chores**
  * Strengthened Helm chart validation in the CI/CD pipeline
  * Updated to Helm v3.20.1 with pinned version dependencies for improved consistency

PR #310, opened 2026-06-18
:bug: helm: fix chart rendering with shared image helpers

kahirokunn:
## Summary
- Split the critical Helm chart rendering fixes out of #306 so they can be reviewed and merged first.
- Add shared image helpers for the top-level cluster-proxy chart and the embedded addon-agent chart.
- Build local Helm chart dependencies in existing build/test/image/release paths so the chart changes work from a clean checkout.
- Leave the new Helm validation target and presubmit job for a follow-up discussion.

## Notes
This intentionally does not include the `test-helm` Makefile target or the dedicated Helm presubmit job from #306.


kahirokunn:

mikeshng (commented):
I am good with merging this. I made a mistake in https://github.com/open-cluster-management-io/cluster-proxy/pull/306#pullrequestreview-4516777540 I thought there was a bug so that's why I asked you fix the bug first and then add the validation. It seems like this PR is about making sure we use common code as much as possible and then the important validation PR will follow up. 

The validation of the embedded manifest/helmChart will be a huge improvement.

Thank you!

This looks good to me if @tesshuflower agrees with adding the Helm dependency (which I think is fine because we are embedding chart(s) under `pkg` so we need to use Helm to validate etc.

/assign @tesshuflower

kahirokunn:
Yes, it is due to standardization and refactoring. This time, I referred to the configuration of the Bitnami Helm chart 👍

tesshuflower:
@kahirokunn this is very cool, but I'm a bit worried about making this change - this would introduce dependency with the bundled charts under manifests that require us to render the helm charts with `helm dependency build` - this is not obvious when you look at the Dockerfile for example.  Someone building the dockerfile alone could miss this, and then we have a silent failure where the bundled manifests are not correct.

I know you updated makefiles etc, but I think this is adding a lot of complexity to understand exactly what gets packaged.

If we do go with this complex dependency loading, I would perhaps prefer that we commit the built charts into the `manifests` directory, rather than dynamically building them at build time.  Then CI checks could confirm by re-running helm dependency build and check that the files match.

However, I'm still not convinced we need this change - This is listed as a bug - is there a specific fix this is addressing with the helm charts or is it mostly about adding some additional helm chart linting and organization?  @mikeshng thoughts?

mikeshng:
I thought there was a bug so I asked Kahiro to split it. It really should be with the validation. As I mentioned in https://github.com/open-cluster-management-io/cluster-proxy/pull/310#pullrequestreview-4525875376 the validation is the net gain. 

We need to validate as best as possible especially the embedded chart because if there is something wrong with that chart, we will have to make code changes to fix it, do a release etc.

If there is something wrong with the non embedded chart, it's much easier to workaround so the real gain is the ability to validate the embedded chart.

kahirokunn:
> I thought there was a bug so I asked Kahiro to split it. It really should be with the validation. As I mentioned in [#310 (review)](https://github.com/open-cluster-management-io/cluster-proxy/pull/310#pullrequestreview-4525875376) the validation is the net gain.
> 
> We need to validate as best as possible especially the embedded chart because if there is something wrong with that chart, we will have to make code changes to fix it, do a release etc.
> 
> If there is something wrong with the non embedded chart, it's much easier to workaround so the real gain is the ability to validate the embedded chart.

@mikeshng  has already expressed everything I wanted to say.
Standardizing, depending on stable components, and conducting unit tests are extremely important in software development.
I believe this PR will further promote those practices.

kahirokunn:
And of course, I have also included some minor bug fixes. Specifically, there was a bug where the Helm chart provided to users could not be deployed using the default values. I do not consider this to be critical. However, when considering why the bug occurred and how to prevent similar issues in the future, I realized it was due to poor readability, a violation of the DRY principle, and a lack of testing. This PR is the result of my efforts to make those improvements.

tesshuflower:
> And of course, I have also included some minor bug fixes. Specifically, there was a bug where the Helm chart provided to users could not be deployed using the default values. I do not consider this to be critical. However, when considering why the bug occurred and how to prevent similar issues in the future, I realized it was due to poor readability, a violation of the DRY principle, and a lack of testing. This PR is the result of my efforts to make those improvements.

I think all that makes sense - my main concern is simply that we're embedding the helm charts in `pkg/proxyagent/agent/manifests` and now with these updates the charts as-is are not complete in the embedded dir - so you need extra steps - resulting in possibly other errors (that are not found at image build time) if you don't do the pre-build steps before building the image.  The end result seems harder to maintain and understand long-term than some duplication.

kahirokunn:
The helper chart only contains utilities that don't change very often, and I've also added hooks to the make command, so I think image builds should go smoothly without any bottlenecks. What do you think?

kahirokunn:
Alternatively, there is also the option of committing helm chart or tgz file, similar to how you handle the vendor directory.

tesshuflower:
> Alternatively, there is also the option of committing helm chart or tgz file, similar to how you handle the vendor directory.

yes, I mentioned this above (although my comment probably got a bit long)... I would prefer this.  Then the ci checks can still run the helm dependency build steps, and simply verify the files match (i.e. that the PR has done the prereq dependency build steps properly and includes the generated files).

The main advantage is we can more easily see exactly what will be embedded, and lower risk of someone building an image that is missing the embedded files.

kahirokunn:
@tesshuflower Thank you for your review! How about now? 👀

hack/verify-helm-dependencies.sh
```diff
@@ -0,0 +1,176 @@
...
+		return 1
+	fi
+
+	if ! diff -u <("${HELM}" show chart "${source_dir}") <("${HELM}" show chart "${archive}"); then
+		echo "${chart_dir}: dependency archive ${archive} has stale Chart.yaml content from ${repository}" >&2
+		return 1
+	fi
+
+	if ! diff -ru -x Chart.yaml "${source_dir}" "${extract_dir}/${name}"; then
+		echo "${chart_dir}: dependency archive ${archive} is out of sync with ${repository}" >&2
```
tesshuflower, line 82: Is it possible to print something out here to indicate how a user should fix this? We could alternatively try to document somewhere - but might be easiest to mention they should run:

`make sync-helm-dependencies`


I will approve however, this is pretty minor and could be done in a follow-on

tesshuflower:
/lgtm

Thanks a lot for the updates @kahirokunn !

PR #312, opened 2026-06-18
:seedling: pin clusteradm for e2e with tools module

kahirokunn:
## Summary
- Build clusteradm from a dedicated `hack/tools` Go module instead of downloading `install.sh` from the mutable `main` branch.
- Pin clusteradm to `open-cluster-management.io/clusteradm v1.3.1`.
- Add Dependabot Go module updates for the main module and `hack/tools`.

## Related Issue
N/A

## Summary by CodeRabbit

* **Chores**
  * Enabled automated dependency updates through Dependabot with daily check frequency and intelligent grouping.
  * Updated CI/CD workflow to build required tools from source rather than external scripts.
  * Added build tooling infrastructure for managing development dependencies.


.github/dependabot.yml
```diff
@@ -0,0 +1,37 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directories:
+      - "/"
+      - "/hack/tools"
+    schedule:
+      interval: "daily"
+      time: "09:00"
+      timezone: "Asia/Tokyo"
```
mikeshng: nit: what is the standard k8s practice for this type of thing? Does it show bias towards any timezone? I don't mind this being Tokyo time at all just wondering.

mikeshng (commented):
I like this change 👍  because:

1) https://github.com/open-cluster-management-io/cluster-proxy/pull/308#issuecomment-4732193070 @kahirokunn mentioned there are other projects that does the samething and I like following convention/standardization.

2) This addresses the scary `install.sh | bash`. There has been heighten security awareness lately especially around GitHub workflow removing something like this sounds good to me.

3) Technically this method also install the latest clusteradm (with a slight time gap) because the dependbot also updates it at an interval.

If other maintainers are good with this change, we should follow the same practice for the other repos.

.github/dependabot.yml
kahirokunn: I checked a few nearby OCM / Kubernetes SIGs repos, and they generally don't specify `time` / `timezone` for Dependabot schedules:

- https://github.com/open-cluster-management-io/managed-serviceaccount/blob/main/.github/dependabot.yml
- https://github.com/open-cluster-management-io/ocm/blob/main/.github/dependabot.yml
- https://github.com/kubernetes-sigs/controller-runtime/blob/main/.github/dependabot.yml
- https://github.com/kubernetes-sigs/cluster-api/blob/main/.github/dependabot.yaml

So I removed the explicit `Asia/Tokyo` setting here too.

tesshuflower:
/lgtm

I was actually originally pushing back on this thinking getting clusteradm latest meant it pulled its images with the `latest` tag, but Mike has corrected me - and the clusteradm install.sh script that was originally used will just pull the latest release (which is still v1.3.1 atm).

So, this is very nice in the end and allows us to stay current via dependabot updates 🚀   Thanks @kahirokunn !

PR #320, opened 2026-06-22
:seedling: Use OCM addon v1beta1

kahirokunn:
## Summary
- migrate OCM addon resources in the cluster-proxy chart from addon v1alpha1 to v1beta1
- add a v1beta1 AddOnDeploymentConfig for the default agent install namespace
- update samples to use AddOnDeploymentConfig and ManagedClusterAddOn spec.configs
- keep the cluster-proxy ManagedProxyConfiguration API at proxy v1alpha1

## Summary by CodeRabbit

* **Bug Fixes**
  * Added a readiness probe to the service-proxy container (TCP check on port 7443) when service proxy is enabled.
* **Documentation**
  * Updated the Helm chart value description for the agent installation namespace.
* **Chores**
  * Upgraded cluster proxy add-on manifests and samples to the v1beta1 API.
  * Updated Helm chart configuration wiring so the manager and user deployments install agents into the configured spoke add-on namespace.
* **Tests**
  * Extended agent tests to cover service-proxy behavior and expected generated manifests.

### DIFF
--- a/charts/cluster-proxy/README.md
+++ b/charts/cluster-proxy/README.md
@@ -26,7 +26,7 @@ helm install cluster-proxy ./charts/cluster-proxy \
 | `image`                                 | Image name                         | `cluster-proxy`                                 |
 | `tag`                                   | Image tag                          | Chart version                                   |
 | `replicas`                              | Number of replicas                 | `1`                                             |
-| `spokeAddonNamespace`                   | Namespace for spoke addon          | `open-cluster-management-cluster-proxy`         |
+| `spokeAddonNamespace`                   | Default agent install namespace    | `open-cluster-management-cluster-proxy`         |
 | `proxyServerImage`                      | Proxy server image                 | `quay.io/open-cluster-management/cluster-proxy` |
 | `proxyAgentImage`                       | Proxy agent image                  | `quay.io/open-cluster-management/cluster-proxy` |
 | `proxyServer.entrypointLoadBalancer`    | Enable LoadBalancer for entrypoint | `false`                                         |
--- a/charts/cluster-proxy/templates/user-deployment.yaml
+++ b/charts/cluster-proxy/templates/user-deployment.yaml
@@ -59,7 +59,7 @@ spec:
             - --server-key=/user-tls/tls.key
             - --server-cert=/user-tls/tls.crt
             - --service-proxy-ca-cert=/proxy-ca/ca.crt
-            - --agent-install-namespace=open-cluster-management-agent-addon
+            - --agent-install-namespace={{ .Values.spokeAddonNamespace }}
           env:
             - name: POD_NAMESPACE
               valueFrom:
--- a/charts/cluster-proxy/templates/clustermanagementaddon.yaml
+++ b/charts/cluster-proxy/templates/clustermanagementaddon.yaml
@@ -1,4 +1,4 @@
-apiVersion: addon.open-cluster-management.io/v1alpha1
+apiVersion: addon.open-cluster-management.io/v1beta1
 kind: ClusterManagementAddOn
 metadata:
   name: cluster-proxy
@@ -8,13 +8,10 @@ spec:
   addOnMeta:
     displayName: cluster-proxy
     description: cluster-proxy
-  supportedConfigs:
+  defaultConfigs:
   - group: proxy.open-cluster-management.io
     resource: managedproxyconfigurations
-    defaultConfig:
-      name: cluster-proxy
-  - group: addon.open-cluster-management.io
-    resource: addondeploymentconfigs
+    name: cluster-proxy
   installStrategy:
     type: Placements
     placements:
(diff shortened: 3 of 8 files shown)


kahirokunn:

charts/cluster-proxy/templates/user-deployment.yaml
```diff
@@ -59,7 +59,7 @@ spec:
             - --server-key=/user-tls/tls.key
             - --server-cert=/user-tls/tls.crt
             - --service-proxy-ca-cert=/proxy-ca/ca.crt
-            - --agent-install-namespace=open-cluster-management-agent-addon
```
mikeshng, line 62: `open-cluster-management-agent-addon` This is the ns we actually want no?

charts/cluster-proxy/README.md
```diff
@@ -26,7 +26,7 @@ helm install cluster-proxy ./charts/cluster-proxy \
 | `image`                                 | Image name                         | `cluster-proxy`                                 |
 | `tag`                                   | Image tag                          | Chart version                                   |
 | `replicas`                              | Number of replicas                 | `1`                                             |
-| `spokeAddonNamespace`                   | Namespace for spoke addon          | `open-cluster-management-cluster-proxy`         |
+| `spokeAddonNamespace`                   | Agent installation namespace       | `open-cluster-management-cluster-proxy`         |
```
mikeshng: I am not an expert of this project but from what I am used to the ocm agents should land in `open-cluster-management-agent-addon` ns. 

What is the default ns currently if I install via clusteradm and what is the default ns after this PR is merged? I want it to be as the same as before `open-cluster-management-agent-addon` if that's still the case.

charts/cluster-proxy/templates/user-deployment.yaml
mikeshng: Maybe this param is not needed anymore because it's deprecated so this removal is still valid but I am mainly concern about the default ns of the agent on the spoke/managed cluster.

charts/cluster-proxy/README.md
tesshuflower: I agree @mikeshng , I had just started to take a look at this as well.

I think generally the namespace in the chart vlaues is essentially what is used as the "default" if no addondeploymentconfig is used to override it.  Which I think is fine to make it configurable, but as Mike is saying we'll just need to make sure it matches what the default was previously.

One thing we do when using these charts is to allow users to override the addondeploymentconfig used - so by setting a default in the ClusterManagementAddOn will stop this from working.  I.e. we may keep pushing our clustermanagementaddon down, which will hardcode a default here:

https://github.com/open-cluster-management-io/cluster-proxy/pull/320/changes#diff-89537e066c7a67567475405903ff7f6f4edb2248d3d1aef5c0b39cc0d9dd23a0R15-R18

What do you think @kahirokunn ?  I think it would be best if we allow setting the default value in the chart values (as you are doing here), however we do not set a default  addondeploymentconfig to allow this to be overridden by users on a case-by base basis.

kahirokunn: Thanks for pointing this out. I agree that `open-cluster-management-agent-addon` is the default namespace used by the generic addon-framework / clusteradm path.

For cluster-proxy specifically, though, the Helm chart default was already `open-cluster-management-cluster-proxy` before this PR, as defined here:
https://github.com/open-cluster-management-io/cluster-proxy/blob/main/charts/cluster-proxy/values.yaml#L13

So this PR does not change the default agent install namespace for the Helm chart install path. The intent is to preserve the existing cluster-proxy default and make the v1beta1 `AddOnDeploymentConfig`, manager flag, and user-server flag all use the same chart value instead of leaving the user-server hardcoded to `open-cluster-management-agent-addon`.

For the clusteradm path, the generic `clusteradm addon enable` default is still `open-cluster-management-agent-addon`, and clusteradm writes that value to the deprecated v1alpha1 `ManagedClusterAddOn.spec.installNamespace`. For cluster-proxy specifically, though, the effective default has already been `open-cluster-management-cluster-proxy` through the cluster-proxy default configuration/chart value, and this PR does not change that behavior.

Changing the Helm default from the existing `open-cluster-management-cluster-proxy` to `open-cluster-management-agent-addon` would be a separate behavior change for users who install the chart with default values.
That is why I think keeping the current chart default is the safer path here.

charts/cluster-proxy/templates/user-deployment.yaml
kahirokunn: same: https://github.com/open-cluster-management-io/cluster-proxy/pull/320#discussion_r3453111348
🙇

charts/cluster-proxy/README.md
mikeshng: > That is why I think keeping the current chart default is the safer path here.

SGTM Leaving it to @tesshuflower

kahirokunn: https://github.com/open-cluster-management-io/cluster-proxy/pull/320#discussion_r3453093524

I'm Reading this comment now... 👀 🙏 🙇

kahirokunn: @tesshuflower Thank you for pointing that out! I've fixed it🙏

tesshuflower:
I think this looks good now, thanks @kahirokunn for these changes!

/lgtm

kahirokunn:
Thank you too 🙏

kahirokunn:
I fixed the CI and squashed the commits into a single commit.

PR #321, opened 2026-06-22
:seedling: Ignore hack tool binaries

kahirokunn:
Summary:
- Add `/hack/tools/bin/` to `.gitignore` so local hack tool binaries are not tracked.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@
 
 /apiserver.local.config
 /bin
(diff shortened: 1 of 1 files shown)


kahirokunn:

mikeshng (approved):
/lgtm

Yes, this is a good fix so we don't accidentally commit binaries. 

THANK YOU!

PR #330, opened 2026-07-21
:seedling: remove duplicate service domain test case

kahirokunn:
## Summary

Remove the second copy of the `TestNewAgentAddon` case that verifies a customized `serviceDomain`.

### DIFF
--- a/pkg/proxyagent/agent/agent_test.go
+++ b/pkg/proxyagent/agent/agent_test.go
@@ -648,28 +648,6 @@ func TestNewAgentAddon(t *testing.T) {
 			},
 		},
 		{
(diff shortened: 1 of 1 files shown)


kahirokunn:
@coderabbitai review

kahirokunn:
@coderabbitai rate limit

mikeshng (approved):
/lgtm

Yes, this is the same exact test in https://github.com/open-cluster-management-io/cluster-proxy/blob/main/pkg/proxyagent/agent/agent_test.go#L482 
removing it make sense. Thank you for finding this and fixing it!

PR #331, opened 2026-07-21
:seedling: use test-scoped HTTP request contexts

kahirokunn:
## Summary

Use test-scoped contexts for HTTP requests in the existing service-proxy unit tests and Ginkgo connectivity tests.


mikeshng (approved):
/lgtm

Great enhancement to improve testing. I don't see any reason to hold this up since the chance of regression is unlikely.

Thank you!

PR #334, opened 2026-07-23
:sparkles: Add opt-in NetworkPolicies for hub and managed workloads

rokej:
## Summary
- Adds opt-in Helm `networkPolicies.enabled` (default `false`) that renders NetworkPolicies for hub addon-manager, user-server (when `enableServiceProxy`), and ANP proxy-server.
- Passes the same toggle to the addon-manager as `--enable-network-policies`, which injects agent chart values so the managed-cluster proxy-agent NetworkPolicy is included in ManifestWorks.
- Policies are **vendor-neutral**: peers are port-based (empty `from`/`to` where needed) so the same manifests work across Kubernetes distributions. Hub and spoke policies share that portable style (DNS :53/:5353, API :443/:6443, service ports). No OpenShift-specific selectors (e.g. `openshift-dns`, ingress `policy-group`).
- When service-proxy is also enabled, the agent NetworkPolicy adds an allow-all egress exemption so arbitrary `Cluster-Proxy-Port` backends are not blocked.

## Test plan
- [x] `go test ./pkg/proxyagent/agent/`
- [x] `helm template` with `networkPolicies.enabled=true` shows hub NPs and `--enable-network-policies=true` on the manager
- [x] Install/upgrade chart with `networkPolicies.enabled=true` and confirm hub NPs are created
- [x] Confirm managed-cluster proxy-agent NetworkPolicy appears after addon reconcile
- [x] Leave default `false` and confirm no NetworkPolicies are rendered/deployed

### DIFF
--- a/charts/cluster-proxy/templates/cluster-proxy-addon-manager-networkpolicy.yaml
+++ b/charts/cluster-proxy/templates/cluster-proxy-addon-manager-networkpolicy.yaml
@@ -1,0 +1,34 @@
+{{- if .Values.networkPolicies.enabled }}
+# Purpose: Default-deny for cluster-proxy addon-manager, then allow DNS and
+# Kubernetes API egress. Manager has no Service/ingress.
+# Peers are port-based (empty "to") for portability across Kubernetes vendors.
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: cluster-proxy-addon-manager-network-policy
+  namespace: {{ .Release.Namespace }}
+spec:
+  podSelector:
+    matchLabels:
+      component: cluster-proxy-manager
+  policyTypes:
+  - Ingress
+  - Egress
+  egress:
+  # DNS (:53 common; :5353 used by some distributions)
+  - ports:
+    - protocol: UDP
+      port: 53
+    - protocol: TCP
+      port: 53
+    - protocol: UDP
+      port: 5353
+    - protocol: TCP
+      port: 5353
+  # Kubernetes API
+  - ports:
+    - protocol: TCP
+      port: 443
+    - protocol: TCP
+      port: 6443
+{{- end }}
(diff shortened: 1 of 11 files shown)


charts/cluster-proxy/templates/cluster-proxy-addon-manager-networkpolicy.yaml
```diff
@@ -0,0 +1,30 @@
...
+  - to:
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: openshift-dns
+    ports:
+    - protocol: TCP
+      port: 5353
+    - protocol: UDP
+      port: 5353
+  - {}
```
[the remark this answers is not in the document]
rokej: They are intentional. kube-apiserver is typically host-networked, so its address/port are not reliably selectable with a namespaced NetworkPolicy. That guideline’s API sample is exactly egress: - {} until something like AdminNetworkPolicy (nodes peer / dynamic endpoints) is used.

charts/cluster-proxy/templates/cluster-proxy-addon-manager-networkpolicy.yaml
```diff
@@ -0,0 +1,30 @@
...
+    matchLabels:
+      component: cluster-proxy-manager
+  policyTypes:
+  - Ingress
+  - Egress
+  egress:
+  - to:
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: openshift-dns
```
mikeshng: If this is OCP only thing then the toggle should specify it. WDYT? CC @tesshuflower 

Maybe `networkPolicies.ocp.enabled` ? or turn it around like `ocp.networkPolicies.enabled` so there might be more ocp stuff in the future?

Same feedback for the other changes below, cmd params, other helm toggles etc.

Otherwise, I think the PR is good because everything is optional and no API change so impossible to cause a regression unless user specify choose this path.

rokej: @mikeshng Thanks for reviewing this. You are correct.

Hub NetworkPolicies are now vendor-neutral.

**Removed**

- `openshift-dns` namespaceSelector
- OpenShift router `policy-group.network.openshift.io/ingress`
- Allow-all egress {}

**Replaced with (same style as the spoke agent)**

- DNS: ports 53 and 5353, empty `to`
- API: 443 / 6443, empty `to`
- Ingress: empty `from` on service ports (works with any ingress controller)
- User ↔ proxy-server: OCM pod label peers only (not platform-specific)

mikeshng: Please update the PR description for the cluster-proxy maintainers to review. Thanks.

PR #335, opened 2026-07-24
:bug: strip client Impersonate-* headers in service-proxy

rokej:
## Summary
- Strip client-supplied `Impersonate-*` headers at the start of `processAuthentication`.
- Re-apply hub-user impersonation headers only from TokenReview results.

## Test plan
- [x] `go test ./pkg/serviceproxy/`
- [x] Manual verification on hub/spoke with a low-privileged hub user

### DIFF
--- a/pkg/serviceproxy/service_proxy.go
+++ b/pkg/serviceproxy/service_proxy.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	"github.com/spf13/cobra"
+	authenticationv1 "k8s.io/api/authentication/v1"
 	"k8s.io/apiserver/pkg/authentication/authenticator"
 	"k8s.io/apiserver/pkg/authentication/token/cache"
 	"k8s.io/apiserver/pkg/authentication/user"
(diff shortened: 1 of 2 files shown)


mikeshng (approved):
This fix looks good to me but will need cluster-proxy maintainer to approve.

/lgtm

/assign @kahirokunn @qiujian16 @tesshuflower

kahirokunn:
/approve
Thank you 🙏

mikeshng (commented):
@qiujian16 @tesshuflower coming back to this PR. It always remove the impersonate header doesn't that mean we might cause a regression for some use cases?

kahirokunn:
For example, if someone has a configuration like kube-oidc-proxy -> cluster-proxy, this PR will break that functionality.
https://github.com/open-cluster-management-io/cluster-proxy/pull/336

qiujian16:
I think instead of always dropping the impersonation header, you might consider when there are impersonation headers from client, the proxy agent should not impersonate using its own credential any longer. It should just forward the request transparently.

kahirokunn:
> I think instead of always dropping the impersonation header, you might consider when there are impersonation headers from client, the proxy agent should not impersonate using its own credential any longer. It should just forward the request transparently.

https://github.com/open-cluster-management-io/cluster-proxy/pull/336
That's certainly true. I've adjusted the PR accordingly.

PR #337, opened 2026-07-26
:bug: recognize the kube-apiserver target by Service identity

kahirokunn:
The service proxy recognized kube-apiserver traffic only when the scheme and port were exactly `https` and `443`. A request such as `https:kubernetes:0443` still reached the API server but followed the regular Service path, skipping authentication and impersonation handling.

Identify the kube-apiserver by the `default/kubernetes` Service and normalize the target URL to `https://kubernetes.default.svc`.

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -18,6 +18,11 @@ const (
 	HEADERSERVICECERT = "Service-Client-Cert"
 	HEADERSERVICEKEY  = "Service-Client-Key"
 
+	KubeAPIServerHost = "kubernetes.default.svc"
+
+	kubeAPIServerService   = "kubernetes"
+	kubeAPIServerNamespace = "default"
+
 	// Cluster-Proxy custom headers for service proxy
 	HeaderClusterProxyProto     = "Cluster-Proxy-Proto"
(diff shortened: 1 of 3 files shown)


kahirokunn:

pkg/utils/utils.go
```diff
@@ -123,8 +128,8 @@ func GetTargetServiceURLFromRequest(req *http.Request) (*url.URL, error) {
 
 	var targetServiceURL string
 	// check if the request is meant to proxy to kube-apiserver
-	if proto == "https" && service == "kubernetes" && namespace == "default" && port == "443" {
-		targetServiceURL = "https://kubernetes.default.svc"
+	if service == kubeAPIServerService && namespace == kubeAPIServerNamespace {
```
mikeshng: These are case sensitive checks so to be triple safe (what this PR is trying to enhance upon) should use case insensitive checks. Otherwise, I think it's LGTM.

kahirokunn: Thank you. Fixed 🙏

mikeshng (approved):
Thank you for addressing the review feedback.

/lgtm

PR #338, opened 2026-07-26
:bug: reuse the outbound transport in service-proxy

kahirokunn:
`ServeHTTP` creates a new `http.Transport` for every request, preventing connection reuse between proxied requests.

Create the transport once during `Run`, reuse it for all requests, and close idle connections when the service proxy stops.


kahirokunn:

mikeshng (approved):
/lgtm

This is a nice catch! Thank you!

mikeshng:
CC @tesshuflower this might help with performance.

PR #339, opened 2026-07-26
:bug: drain in-flight requests on shutdown

kahirokunn:
Some proxy shutdown paths call `klog.Fatal` or `os.Exit`, so TLS certificate rotation can interrupt active requests and upgraded connections such as `kubectl exec` and `port-forward`.

Route shutdown through context cancellation, drain HTTP servers within a timeout, force-close hijacked connections at the deadline, and give the pods enough termination grace time to finish.

### DIFF
--- a/charts/cluster-proxy/templates/manager-deployment.yaml
+++ b/charts/cluster-proxy/templates/manager-deployment.yaml
@@ -19,6 +19,9 @@ spec:
         component: cluster-proxy-manager
     spec:
       serviceAccount: cluster-proxy
+      # Binaries drain for up to 30s after SIGTERM; the extra 10s lets forced
+      # cleanup finish before the kubelet sends SIGKILL.
+      terminationGracePeriodSeconds: 40
       securityContext:
         runAsNonRoot: true
         seccompProfile:
--- a/pkg/controllers/manager.go
+++ b/pkg/controllers/manager.go
@@ -2,7 +2,7 @@ package controllers
 
 import (
 	"context"
-	"os"
+	"fmt"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -17,7 +17,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
-	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 
 	"open-cluster-management.io/cluster-proxy/pkg/proxyserver/operator/authentication/selfsigned"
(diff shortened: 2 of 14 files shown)


kahirokunn:

charts/cluster-proxy/templates/manager-deployment.yaml
```diff
@@ -19,6 +19,8 @@ spec:
         component: cluster-proxy-manager
     spec:
       serviceAccount: cluster-proxy
+      # Allow the 30s shutdown timeout to complete.
+      terminationGracePeriodSeconds: 40
```
mikeshng, line 24: is it 40 or 30 the comment says 30s but the value is 40? Same feedback for below.

pkg/controllers/manager.go
```diff
@@ -66,24 +62,25 @@ func addFlags(cmd *cobra.Command) {
...
 	nativeClient, err := kubernetes.NewForConfig(kubeConfig)
 	if err != nil {
-		return err
+		return fmt.Errorf("create Kubernetes client: %w", err)
 	}
 	secertClient := nativeClient.CoreV1()
 	informerFactory := informers.NewSharedInformerFactory(nativeClient, 10*time.Minute)
 	secertLister := informerFactory.Core().V1().Secrets().Lister()
 
-	go informerFactory.Start(ctx.Done())
```
mikeshng, line 85: Why did you move this? To go along with the whole design of the PR to be sync as much as possible? I prefer if you don't touch this tho because I feel like it's not in the scope for this PR unless I misunderstand it.

mikeshng (commented):
So I gave it a shot at reviewing this. This PR title is a bit deceptive of the significant amount of changes involved. I do agree this is a bug and needs to be addressed for sure. I need someone else to take a look at this for sure because

- it introduces a new http server wrapper/framework that it leverages 
- it changes how the binary start and stop (maybe rightfully so but still scary)

I guess the fix is this involved because there is no other way around it and it's very tricky to deal with something like this.

/assign @tesshuflower @qiujian16

charts/cluster-proxy/templates/manager-deployment.yaml
kahirokunn: The comment was misleading, sorry. 40 is on purpose: after SIGTERM the binaries drain in-flight work for up to 30s (DefaultHTTPShutdownTimeout), and the extra 10s is headroom so the forced cleanup and process exit can finish before the kubelet sends SIGKILL. If the pod grace period were also 30s we'd get killed right in the middle of the drain. I reworded the comment in all three deployments to actually say that.

pkg/controllers/manager.go
kahirokunn: No strong reason, so I reverted it. informerFactory.Start doesn't block (it spawns goroutines internally), so the go was redundant, and I wanted the informers to start only after all the setup that can fail. But none of that is needed for the shutdown fix, so it's back where it was.

kahirokunn:
Splitting this up since the single PR was carrying too much at once:

- #342 restarts the hub manager gracefully on TLS changes
- #343 returns errors instead of exiting mid-setup
- #344 is the actual shutdown draining, stacked on #343

The inline comments are addressed there: the informer startup move is reverted, and the grace period comments now spell out the 40 (30s in-process drain plus 10s headroom before SIGKILL). Closing this one in favor of the stack. Thanks for the review @mikeshng!

PR #342, opened 2026-08-04
:bug: restart hub manager gracefully on TLS changes

kahirokunn:
Fixes the addon-manager killing itself with `os.Exit(0)` on TLS ConfigMap changes, which skipped graceful shutdown.

### DIFF
--- a/cmd/addon-manager/main.go
+++ b/cmd/addon-manager/main.go
@@ -183,8 +183,8 @@ func main() {
 	// So we could alternatively add a watch there on the configmap and simply reconcile.
 	// This way keeps it simpler with common code, but does restart when the configmap changes
 	sdkTLSConfig, err := sdktls.StartTLSConfigMapWatcher(ctx, nativeClient, podNamespace, func() {
-		klog.Info("TLS ConfigMap changed, restarting")
-		os.Exit(0)
+		klog.Info("TLS ConfigMap changed, shutting down gracefully for restart")
+		cancel()
 	})
 	if err != nil {
 		setupLog.Error(err, "failed to start TLS ConfigMap watcher")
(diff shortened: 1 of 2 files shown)


kahirokunn:
@mikeshng @tesshuflower  Hi ✋ PTAL. Thank you 🙏

cmd/addon-manager/main.go
```diff
@@ -183,8 +183,8 @@ func main() {
 	// So we could alternatively add a watch there on the configmap and simply reconcile.
 	// This way keeps it simpler with common code, but does restart when the configmap changes
 	sdkTLSConfig, err := sdktls.StartTLSConfigMapWatcher(ctx, nativeClient, podNamespace, func() {
-		klog.Info("TLS ConfigMap changed, restarting")
-		os.Exit(0)
+		klog.Info("TLS ConfigMap changed, shutting down gracefully for restart")
+		cancel()
```
mikeshng, line 187: @tesshuflower please make sure this is going to be ok for downstream.

Makefile
```diff
@@ -140,6 +140,13 @@ test-helm: verify-helm-dependencies validate-values-schema ## Lint and render He
 		--namespace open-cluster-management-agent-addon \
 		--set oidcReservedNamePrefixes=null 2>&1; } | grep -q oidcReservedNamePrefixes
 
+.PHONY: verify-shutdown-grace-periods
+verify-shutdown-grace-periods:
```
mikeshng: Do we really need this? It seems like an overkill? We don't check all the other values, why do we place such an importance on this specific value? If it's that important, maybe it should be in the README.

kahirokunn: Sorry, I've deleted it.

kahirokunn:
@tesshuflower For reference, I’ve recently noticed something: cluster-proxy appears to be designed for on-demand instances where it’s expected to be rarely used.
However, to maximize cloud resource utilization, I’m aggressively rebalancing pod placement primarily using spot instances with Karpenter to optimize resources.
Therefore, I want to stabilize cluster-proxy’s startup/shutdown processes.

tesshuflower:
thanks for the added context @kahirokunn! 👍 

/lgtm

kahirokunn:
Thank you too 🙌

PR #343, opened 2026-08-04
:seedling: return errors instead of exiting mid-setup

kahirokunn:
The setup paths in the cluster-proxy binaries exit through `klog.Fatal` / `os.Exit`, which skips deferred cleanup and the log flush (golangci's exitAfterDefer flags two of these on main today). This converts them to returned errors and lets `main` print the error and pick the exit code. The root command now requires a subcommand via `RunE`.

No signal handling is touched and there's no behavior change beyond the error output. Groundwork for the shutdown draining change.

Split out of open-cluster-management-io/cluster-proxy#339.

### DIFF
--- a/cmd/cluster-proxy/main.go
+++ b/cmd/cluster-proxy/main.go
@@ -1,8 +1,10 @@
 package main
 
 import (
+	"errors"
 	goflag "flag"
 	"fmt"
+	"io"
 	"os"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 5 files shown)


kahirokunn:

cmd/cluster-proxy/main.go
```diff
@@ -24,24 +26,37 @@ func main() {
 	pflag.CommandLine.AddGoFlagSet(goflag.CommandLine)
 
 	logs.InitLogs()
-	defer logs.FlushLogs()
+	os.Exit(runMain(execute, os.Stderr, logs.FlushLogs))
```
mikeshng, line 29: The title of the PR and commit msg should be updated/changed. We are still exiting, but just doing cleanup first then exit.

PR #344, opened 2026-08-04
:sparkles: drain proxy HTTP requests on shutdown

kahirokunn:
The user-server and service-proxy currently stop without giving active HTTP requests time to finish when they receive SIGTERM or reload their TLS configuration. This can interrupt proxy traffic during rollouts, node drains, preemption, or a TLS profile change.

This change adds a readiness-first shutdown sequence around the public HTTP servers. Each server reports itself unready, keeps its listener available for a short endpoint propagation window, and then uses Go's standard [`http.Server.Shutdown`](https://pkg.go.dev/net/http#Server.Shutdown) to close the listener and idle connections while active HTTP requests finish within the drain timeout.

Connections hijacked from `net/http`, including WebSocket- and SPDY-based exec, attach, and port-forward sessions, are intentionally not included in the drain. Waiting could allow some of those sessions to close naturally, but cluster-proxy cannot send a common shutdown notification or transfer them to another Pod. They are disconnected when the process exits. Keeping this boundary lets the implementation use the standard HTTP lifecycle without a custom TCP connection tracker.

On a TLS configuration change, the affected container follows the same shutdown sequence and exits. The kubelet restarts that container in the existing Pod so the new TLS settings are loaded.

## Shutdown flow

```mermaid
sequenceDiagram
    participant S as Shutdown trigger
    participant P as Proxy process
    participant K as Kubernetes
    participant C as Existing clients

    S->>P: SIGTERM or TLS configuration change
    P->>P: Mark /readyz unready
    K->>P: Probe /readyz
    P-->>K: 503 Service Unavailable
    K->>K: Remove the Pod from endpoints

    Note over P,K: Keep the listener open for the 10s propagation window

    P->>P: Call http.Server.Shutdown
    P->>P: Close the listener and idle HTTP connections
    P-->>C: Allow active HTTP requests to finish

    alt Active HTTP requests finish
        P->>P: Exit
    else 30s drain timeout expires
        P->>P: Exit with active requests
    end

    Note over P,C: Hijacked upgrade connections are not awaited and disconnect on exit
```

## What changed

- Add separate `/healthz` and `/readyz` endpoints.
- Drain HTTP requests with `http.Server.Shutdown` after an endpoint propagation delay.
- Replace `os.Exit(0)` on TLS changes with context cancellation so the same shutdown path runs first.
- Route process signals through the root command context.
- Use controller-runtime's default graceful shutdown for the controllers command.
- Set the Pod termination grace period to 40 seconds.

The user-server and service-proxy expose:

- `--drain-timeout` (default `30s`)
- `--min-drain-duration` (default `10s`)

The minimum duration is included in the overall drain timeout. Service-proxy overrides can be passed through `ManagedProxyConfiguration.spec.proxyAgent.additionalServiceProxyArgs`.

### DIFF
--- a/charts/cluster-proxy/templates/user-deployment.yaml
+++ b/charts/cluster-proxy/templates/user-deployment.yaml
@@ -20,6 +20,9 @@ spec:
         component: cluster-proxy-addon-user
     spec:
       serviceAccount: cluster-proxy
+      # By default, shutdown can take up to 30s after SIGTERM; the extra 10s
+      # lets process cleanup finish before the kubelet sends SIGKILL.
+      terminationGracePeriodSeconds: 40
       securityContext:
         runAsNonRoot: true
         seccompProfile:
@@ -73,6 +76,11 @@ spec:
             privileged: false
             runAsNonRoot: true
             readOnlyRootFilesystem: true
+          readinessProbe:
+            httpGet:
+              path: /readyz
+              port: 8000
+            periodSeconds: 2
           volumeMounts:
             - name: user-tls-vol
               mountPath: /user-tls
--- a/pkg/utils/http_server.go
+++ b/pkg/utils/http_server.go
@@ -1,0 +1,169 @@
+package utils
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+	"sync/atomic"
+	"time"
+
+	"github.com/spf13/pflag"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/healthz"
+)
+
+const (
+	DefaultDrainTimeout     = 30 * time.Second
+	DefaultMinDrainDuration = 10 * time.Second
+)
+
+// DrainConfig controls how long active HTTP requests may remain during shutdown.
+// MinDuration is part of, rather than additional to, Timeout.
+type DrainConfig struct {
+	Timeout     time.Duration
+	MinDuration time.Duration
+}
+
+func (c DrainConfig) Validate() error {
+	if c.Timeout < 0 {
+		return fmt.Errorf("drain-timeout must not be negative")
+	}
+	if c.MinDuration < 0 {
+		return fmt.Errorf("min-drain-duration must not be negative")
+	}
+	if c.MinDuration > c.Timeout {
+		return fmt.Errorf("min-drain-duration must not exceed drain-timeout")
+	}
+	return nil
+}
+
+// AddFlags registers the drain flags shared by the proxy server commands.
+func (c *DrainConfig) AddFlags(flags *pflag.FlagSet) {
+	flags.DurationVar(&c.Timeout, "drain-timeout", DefaultDrainTimeout,
+		"Maximum time to drain active HTTP requests during shutdown.")
+	flags.DurationVar(&c.MinDuration, "min-drain-duration", DefaultMinDrainDuration,
+		"Minimum drain duration allowing time for endpoint deprogramming.")
+}
+
+type HealthProbeServer struct {
+	*http.Server
+	ready atomic.Bool
+}
+
+// NewHealthProbeServer creates separate liveness and readiness endpoints.
+// Readiness only represents whether the proxy is accepting new traffic; the
+// existing config checks remain on the liveness endpoint.
+func NewHealthProbeServer(healthProbeBindAddress string, customChecks ...healthz.Checker) *HealthProbeServer {
+	healthServer := &HealthProbeServer{}
+	healthServer.ready.Store(true)
+
+	mux := http.NewServeMux()
+	checks := map[string]healthz.Checker{
+		"healthz-ping": healthz.Ping,
+	}
+	for i, check := range customChecks {
+		checks[fmt.Sprintf("custom-healthz-checker-%d", i)] = check
+	}
+
+	mux.Handle("/healthz", http.StripPrefix("/healthz", &healthz.Handler{Checks: checks}))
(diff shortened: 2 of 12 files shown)


kahirokunn:
Hi @mikeshng , @tesshuflower , thanks as always! When you get a chance, it'd be great if you could take a look at this one too. 🙏

kahirokunn:
This is based on the same idea.
https://github.com/open-cluster-management-io/cluster-proxy/pull/342#issuecomment-5193405457

tesshuflower:
@kahirokunn this ends up being a very big change - are we sure of the value of this one?  Do we actually expect these pods to get moved around a lot in practice?

the tls Profile case for example should never result in many outtages.

pkg/utils/http_server.go
```diff
@@ -0,0 +1,266 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/healthz"
+)
+
+const (
+	healthProbeReadHeaderTimeout = 5 * time.Second
+	defaultHTTPShutdownTimeout   = 30 * time.Second
+)
+
+// RunnableHTTPServer contains an HTTP server and its serve function.
+type RunnableHTTPServer struct {
```
mikeshng: Is this code copy from somewhere? Can we use a popular library instead? Is it following a convention? We probably need a doc file or at least a README update for this. 

I think following convention is the most important. If we can prove that it's done somewhere else in a popular project then maybe it's justifiable because it does add to future maintainence.

kahirokunn:
@tesshuflower @mikeshng Thanks for the feedback. After thinking through the trade-offs, I simplified the implementation considerably by intentionally not tracking hijacked/upgraded connections, which may be long-lived.

There is no generic server-side mechanism to gracefully terminate and resume those sessions anyway, so tracking and waiting for them would add complexity without providing a reliable graceful-shutdown path. The
implementation now relies on the standard http.Server.Shutdown behavior to drain regular in-flight HTTP requests, while upgraded connections are disconnected when the process exits.

I’d appreciate another review when you have a chance. 🙏

charts/cluster-proxy/templates/user-deployment.yaml
```diff
@@ -73,6 +76,17 @@ spec:
...
             readOnlyRootFilesystem: true
+          livenessProbe:
+            httpGet:
+              path: /healthz
+              port: 8000
+            initialDelaySeconds: 2
+            periodSeconds: 10
+          readinessProbe:
+            httpGet:
+              path: /readyz
```
mikeshng, line 81: The reduced scope makes this PR 10 times better. Thank you for doing that!

This PR makes no changes to the actual proxy logic and the critical request/auth area. 

I will say the shutdown aspect of the PR is good to go. The main goal is complete.

My biggest concern is now do we actually need the probe addition here? My experience is that these probe when failed is quite silent and can cause a huge issue. So on one hand we are handling shutdown better draining stuffs better etc but on the other hand we are introducing this huge risk factor. Let's not make this change unless its necessary which I don't think it is. If it wasn't present before, let's leave it for now and drastically reduce the risk.

CC @tesshuflower

kahirokunn: Thank you for the review! Indeed, the liveness probe is outside the scope of this PR, so I've separated it out.

charts/cluster-proxy/templates/user-deployment.yaml
```diff
@@ -73,6 +76,11 @@ spec:
             privileged: false
             runAsNonRoot: true
             readOnlyRootFilesystem: true
+          readinessProbe:
```
mikeshng, line 79: @kahirokunn this is still here and should removed as discussed

kahirokunn: Could you please take another look at the `Shutdown flow` described in this PR (https://github.com/open-cluster-management-io/cluster-proxy/pull/344#:~:text=settings%20are%20loaded.-,Shutdown%20flow,-What%20changed) and reconsider keeping just the readiness probe? Probes are an important part of a Pod’s lifecycle. I agree that they can introduce operational risk if configured incorrectly, but when handled carefully, they can significantly improve stability. In this design, the readiness probe removes the draining Pod from Service endpoints so that new traffic is routed to other Pods. Together with the drain sequence, this substantially improves stability during scale-in, including when replicas are managed by HPA or KEDA.

mikeshng: I agree with you.

mikeshng (approved):
@tesshuflower I gave this PR a really detailed review. To summaries again:

- The reduced scope makes this PR 10 times better and safer.
- This PR makes no changes to the actual proxy logic and the critical request/auth area.
- I will say the shutdown aspect of the PR is good to go. The main goal is complete.

Please review. I will let it sit until Tuesday morning EST and I will merge it. Thanks!

/lgtm

/hold

tesshuflower:
@kahirokunn thanks for making these changes, this is much simpler to follow (and I think to maintain going forwards).

@mikeshng I know you had issues with modifying the probes, which is understandable - however I think we'll be ok with this in the latest iteration - essentially the readyz probe will move from TCP on the port to simply making sure the http server is up so we will need to adjust accordingly in our downstream charts - however I think the tradeoff is good as when a shutdown happens kube will stop routing to the pod very quickly while the shutdown can continue.  What do you think? I'm ok to lgtm this in the current state but please let me know if you have any concerns.

tesshuflower:
thanks Mike!

/lgtm

kahirokunn:
Thank you 🙏

PR #345, opened 2026-08-05
:sparkles: set a timeout for proxy request headers

kahirokunn:
Proxy listeners currently allow clients unlimited time to send request headers. Set a 32-second header timeout, matching the [Kubernetes API server's secure serving default](https://github.com/kubernetes/kubernetes/blob/v1.36.3/staging/src/k8s.io/apiserver/pkg/server/secure_serving.go#L168-L176).


kahirokunn:

mikeshng (approved):
I was going to suggest make this configurable but it seems like k8s core also don't make this configurable. So I rather we follow convention. This is good to go. Thank you for this enhancement!

/lgtm

PR #346, opened 2026-08-05
🌱 refactor(serviceproxy): introduce authentication provider abstraction

kahirokunn:
## Summary

- introduce a common interface and ordered flow for authentication providers
- move hub and OIDC flags, validation, setup, and config files into provider-specific factories
- allow OIDC authentication when hub token authentication is disabled
- only require the hub kubeconfig when the hub provider is enabled
- keep client `Impersonate-*` requests on the direct pass-through path


kahirokunn:

kahirokunn:
I got the idea after reading this comment from @tesshuflower . 🙏 
https://github.com/open-cluster-management-io/cluster-proxy/pull/336#issuecomment-5182708109

tesshuflower:
sorry @kahirokunn I wasn't actually saying we should rename the flag - I'm not sure we should bother changing this atm? 

This PR also seems to change behaviour wrt OIDC?  I'm not sure of the original intention to make sure the enableImpersonation flag was tied to this.  Do you have more background on why you're making this change?

mikeshng (commented):
Let's not do the rename because there is no real gain. If we want to make OIDC changes let's make this PR about that.

kahirokunn:
@tesshuflower @mikeshng Thanks for clarifying. Some background on the OIDC change: OIDC authentication relies on impersonation internally. Previously, the required impersonation RBAC permissions were granted only when enableImpersonation was enabled, so I initially made OIDC depend on that flag. Given the flag name, that seemed reasonable at the time.

Your earlier comment made me realize that the flag actually controls hub token authentication, rather than impersonation itself. That’s why I renamed it to enableHubTokenAuthentication and separated OIDC from it. I thought this would make its purpose clearer to both users and developers.

I still think the rename is worthwhile because it better reflects what the flag does, but it doesn’t add any functional value. I’ve now removed that refactoring from this PR and kept the changes focused on the OIDC behavior.

I’d appreciate it if you could take another look when you have a chance.

pkg/serviceproxy/service_proxy.go
```diff
@@ -216,40 +219,34 @@ func (s *serviceProxy) Run(ctx context.Context) error {
...
-		// auth — without failure caching, each singleflight group completion triggers a new
-		// API call, causing latency spikes under high concurrency.
-		s.managedClusterAuthenticator = cache.New(managedClusterAuthn, false, s.tokenReviewCacheTTL, s.tokenReviewCacheTTL)
-		s.hubAuthenticator = cache.New(hubAuthn, false, s.tokenReviewCacheTTL, s.tokenReviewCacheTTL)
-		klog.Infof("TokenReview cache enabled with TTL %v", s.tokenReviewCacheTTL)
-	} else {
-		s.managedClusterAuthenticator = managedClusterAuthn
-		s.hubAuthenticator = hubAuthn
-		klog.Infof("TokenReview cache disabled")
+		if s.enableImpersonation {
```
mikeshng: Why do you have to touch this? It feels like a big block of change for cleanup but not related to "decouple OIDC from hub token authentication"

pkg/serviceproxy/service_proxy.go
```diff
@@ -155,19 +155,22 @@ func (s *serviceProxy) Run(ctx context.Context) error {
...
+	if err := s.validate(); err != nil {
+		return err
+	}
+
 	var err error
 	customChecks := []healthz.Checker{}
-
-	cc, err := addonutils.NewConfigChecker("cert", s.cert, s.key, rootCAFile, s.hubKubeConfig)
+	configFiles := []string{s.cert, s.key, rootCAFile}
+	if s.enableImpersonation {
```
mikeshng: Not related to OIDC change? Let's leave it alone?

mikeshng (commented):
@kahirokunn first of all I want to say thank you for all your work including this one. I really think you are doing amazing job.

I started reviewing the code and I think it's just badly structured from the beginning and because the complexity its getting harder and harder to follow which is NOT GOOD for this critical auth area. 

Instead of "patching" it, maybe we have to "redesign" it to address it once. Again **clean code in this area is critical** as we already know from the past few weeks experience. 

How about something like below to clean everything up and clearly state the 3 different cases (and if we ever add more cases in the future it will be easier as well):

```
type authProvider struct {
	name          string
	authenticator authenticator.Token
	process       func(ctx context.Context, req *http.Request, u user.Info) error
}

func (s *serviceProxy) authProviders() []authProvider {
	providers := make([]authProvider, 0, 3)
	if s.managedClusterAuthenticator != nil {
		providers = append(providers, authProvider{name: "managed cluster", authenticator: s.managedClusterAuthenticator})
	}
	if s.hubAuthenticator != nil {
		providers = append(providers, authProvider{name: "hub", authenticator: s.hubAuthenticator, process: s.processHubUser})
	}
	if s.oidcAuthenticator != nil {
		providers = append(providers, authProvider{name: "oidc", authenticator: s.oidcAuthenticator, process: s.processOIDCUser})
	}
	return providers
}

func (s *serviceProxy) processAuthentication(ctx context.Context, req *http.Request) error {
	logger := klog.FromContext(ctx)
	token := strings.TrimPrefix(req.Header.Get("Authorization"), "Bearer ")

	for _, p := range s.authProviders() {
		resp, authenticated, err := p.authenticator.AuthenticateToken(ctx, token)

		if err != nil {
			if !errors.Is(err, ErrTokenNotAuthenticated) {
				return fmt.Errorf("%s authentication failed: %v", p.name, err)
			}
			authenticated = false
		}

		logger.V(4).Info("authentication result", "provider", p.name, "authenticated", authenticated)

		if !authenticated {
			continue // token just isn't valid for this provider, try the next one
		}

		if p.process != nil {
			if err := p.process(ctx, req, resp.User); err != nil {
				return fmt.Errorf("failed to process %s user: %v", p.name, err)
			}
		}
		return nil // success, we're done
	}

	return s.authenticationFailureError() // every configured provider rejected the token
}
...

CC @tesshuflower let's clean this up once and for all for our sanity.

File: pkg/serviceproxy/service_proxy.go (hunk 1 above)
Comment (kahirokunn) [reply pkg/serviceproxy/service_proxy.go]: Yeah, good point. I mixed in some cleanup that wasn't needed for this change. I've removed that version. The current code keeps the cache setup inside the providers, and `Run` just initializes them.

File: pkg/serviceproxy/service_proxy.go (hunk 2 above)
Comment (kahirokunn) [reply pkg/serviceproxy/service_proxy.go]: Agreed. I reverted the unrelated `validate()` move. I kept the config file change so OIDC-only mode does not require a hub kubeconfig.

**Comment by kahirokunn** [comment]
Thanks, this makes sense. I reworked it around a list of providers and a single auth loop, pretty close to what you suggested. I also put each provider's setup in its own factory so we won't have to change the auth flow every time we add one.

File: pkg/serviceproxy/authentication.go [hunk 3]
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// impersonateUser sets the impersonation headers for the given identity and
+// replaces the original token with the cluster-proxy service-account token
+// which has impersonate permission.
+func (s *serviceProxy) impersonateUser(ctx context.Context, req *http.Request, username string, groups []string) error {
+	logger := klog.FromContext(ctx)
+
+	// Ensure no leftover client Impersonate-Group values before adding authenticated groups.
+	req.Header.Del(authenticationv1.ImpersonateGroupHeader)
```
mikeshng: I really like this. This is exactly what we should've done to safeguard even when we always forward the impersonation request. @tesshuflower this is what we want so this is good.

I do remember there were 3 impersonate headers that can be deleted and now you are only deleting the group one. Let's be completely safe and get rid of the other 2 impersonate headers as well.

mikeshng (commented):
Rebase is needed and also a comment about safeguard. More general feedback to come. This is looking very very solid already. Exactly what I wanted!!!

pkg/serviceproxy/authentication.go
kahirokunn: Come to think of it, that's right! I've also added functionality to remove remaining headers!
I've also resolved any conflicts that existed.


```diff
diff --git a/pkg/serviceproxy/authentication.go b/pkg/serviceproxy/authentication.go
index 7af84ad0..873fef04 100644
--- a/pkg/serviceproxy/authentication.go
+++ b/pkg/serviceproxy/authentication.go
@@ -27,18 +27,30 @@ func (s *serviceProxy) readImpersonateTokenFromFile() (string, error) {
 // authenticationv1: Impersonate-User, Impersonate-Group, Impersonate-Uid and Impersonate-Extra-<key>.
 const impersonateHeaderPrefix = "Impersonate-"

+func isImpersonationHeader(key string) bool {
+       return len(key) >= len(impersonateHeaderPrefix) &&
+               strings.EqualFold(key[:len(impersonateHeaderPrefix)], impersonateHeaderPrefix)
+}
+
 // hasClientImpersonationHeaders reports whether the request contains a Kubernetes impersonation
 // header. Matching the complete prefix case-insensitively also covers future impersonation headers.
 func hasClientImpersonationHeaders(headers http.Header) bool {
        for key := range headers {
-               if len(key) >= len(impersonateHeaderPrefix) &&
-                       strings.EqualFold(key[:len(impersonateHeaderPrefix)], impersonateHeaderPrefix) {
+               if isImpersonationHeader(key) {
                        return true
                }
        }
        return false
 }

+func deleteImpersonationHeaders(headers http.Header) {
+       for key := range headers {
+               if isImpersonationHeader(key) {
+                       delete(headers, key)
+               }
+       }
+}
+
 // processAuthentication tries each configured authentication provider in order.
 // Infrastructure errors stop the flow; only an unauthenticated result falls
 // through to the next provider.
@@ -120,10 +132,10 @@ func applyExternalIdentity(ctx context.Context, req *http.Request, externalUser
 func (s *serviceProxy) impersonateUser(ctx context.Context, req *http.Request, username string, groups []string) error {
        logger := klog.FromContext(ctx)

-       // Ensure no leftover client Impersonate-Group values before adding authenticated groups.
-       req.Header.Del(authenticationv1.ImpersonateGroupHeader)
+       // Ensure no client-supplied impersonation values survive when applying the authenticated identity.
+       deleteImpersonationHeaders(req.Header)
        for _, group := range groups {
-               // Add (not Set) so multiple groups are preserved after the Del above.
+               // Add (not Set) so multiple groups are preserved.
                req.Header.Add(authenticationv1.ImpersonateGroupHeader, group)
        }
```

mikeshng (approved):
All feedback has been addressed. The fail safe remove headers is what we want to do. The auth provider abstraction rework is what I wanted. This PR will help with our future maintenance of this critical auth path code. @tesshuflower please review and unhold if you think it's good to go. Thanks.

/hold

/lgtm

tesshuflower:
/lgtm

Thanks for all this @kahirokunn !

/unhold

PR #347, opened 2026-08-05
:seedling: Bump apiserver-network-proxy to v0.36.0

kahirokunn:
`cmd/Dockerfile` still built apiserver-network-proxy v0.35.0 even though the Go modules are already on v0.36.0. This updates the binaries bundled in the image to v0.36.0.


kahirokunn:
Thank you 🙏

PR #348, opened 2026-08-05
✨ Initialize OIDC authentication from configuration events

kahirokunn:
## Why

OIDC settings are delivered before user traffic, but the verifier was initialized lazily by the first bearer-token request. That request therefore had two jobs: start issuer discovery/JWKS loading and authenticate the user. Because discovery is asynchronous, a valid first request could observe the verifier before it was ready and receive a 401; retrying later would then succeed.

A user request should not be the readiness trigger for authentication infrastructure. The goal here is that, once the configured state has converged, the first valid OIDC request only performs authentication. This removes lazy initialization as a cause of first-request 401s without hiding genuine issuer or CA failures.

## Approach

Treat OIDC configuration as desired state and reconcile it independently from request traffic.

```mermaid
sequenceDiagram
    participant A as Add-on configuration
    participant S as service-proxy
    participant I as OIDC issuer
    participant C as Client

    Note over S,C: Before: request-driven initialization
    C->>S: First valid OIDC token
    S->>I: Start discovery and JWKS loading
    S-->>C: 401 while verifier is not ready
    I-->>S: Verifier becomes ready

    Note over A,C: After: configuration-driven initialization
    A->>S: OIDC settings or CA ConfigMap event
    S->>I: Start discovery and JWKS loading
    I-->>S: Verifier becomes ready
    C->>S: First valid OIDC token
    S-->>C: Authenticated request
```

Static settings are reconciled during startup. For ConfigMap-backed CAs, the initial informer state and create, update, and delete events all invoke the same idempotent reconciliation path. This also gives CA correction and rotation a lifecycle that does not require a Pod rollout.

This is intentionally not a request-side retry or a ConfigMap polling loop. Kubernetes already gives us events for desired-state changes, while its OIDC authenticator owns asynchronous retries for issuer discovery. Missing, deleted, or invalid CA data fails OIDC closed and cancels the previous verifier; managed-cluster and hub TokenReview authentication remain available.

## Validation

Verified manually on a kind hub with two managed clusters, using Dex
as the OIDC issuer and real tokens:

- Creating the CA ConfigMap after startup enables OIDC without a pod
  restart or rollout.
- Deleting the ConfigMap, or replacing its data with malformed PEM,
  makes OIDC requests fail with 401 while managed-cluster and hub
  TokenReview authentication keep working.
- Restoring valid CA data recovers OIDC in place.
- Metadata-only ConfigMap updates do not re-trigger initialization.

### DIFF
--- a/pkg/serviceproxy/auth_provider.go
+++ b/pkg/serviceproxy/auth_provider.go
@@ -51,6 +51,7 @@ type authProviderFactoryLogger interface {
 
 type authProviderDependencies struct {
 	managedClusterKubeClient kubernetes.Interface
+	podNamespace             string
 	kubeClientQPS            float32
 	kubeClientBurst          int
 	tokenReviewCacheTTL      time.Duration
@@ -64,9 +65,10 @@ func defaultAuthProviderFactories() []authProviderFactory {
 	}
 }
 
-func (s *serviceProxy) authProviderDependencies() authProviderDependencies {
+func (s *serviceProxy) authProviderDependencies(podNamespace string) authProviderDependencies {
 	return authProviderDependencies{
 		managedClusterKubeClient: s.managedClusterKubeClient,
+		podNamespace:             podNamespace,
 		kubeClientQPS:            s.kubeClientQPS,
 		kubeClientBurst:          s.kubeClientBurst,
 		tokenReviewCacheTTL:      s.tokenReviewCacheTTL,
@@ -74,7 +76,7 @@ func (s *serviceProxy) authProviderDependencies() authProviderDependencies {
 	}
 }
 
-func (s *serviceProxy) initializeAuthProviders(ctx context.Context) error {
+func (s *serviceProxy) initializeAuthProviders(ctx context.Context, podNamespace string) error {
 	enabledFactories := make([]authProviderFactory, 0, len(s.authProviderFactories))
 
 	for _, factory := range s.authProviderFactories {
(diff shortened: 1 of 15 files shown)


kahirokunn:
@mikeshng @tesshuflower Hi ✋ PTAL. Thank you 🙏

pkg/serviceproxy/auth_provider.go
```diff
@@ -64,17 +65,18 @@ func defaultAuthProviderFactories() []authProviderFactory {
 	}
 }
 
-func (s *serviceProxy) authProviderDependencies() authProviderDependencies {
+func (s *serviceProxy) authProviderDependencies(podNamespace string) authProviderDependencies {
 	return authProviderDependencies{
 		managedClusterKubeClient: s.managedClusterKubeClient,
+		podNamespace:             podNamespace,
```
mikeshng, line 71: There is nothing wrong here but it's so awkward coding style and maintainability wise. 

authProviderDependencies is something that all auth providers suppose to use and passing this oidc specific value via function param just feels like a hack.

If it doesn't break anything, I expect this to be set `podNs = s.podNs` and the key initializeAuthProviders function to not have to add this extra podNs param.

mikeshng (approved):
I like this PR because it's only impacting OIDC path. 

The only place I want to critic is the common auth provider maybe a nit but I think it's worth fixing because its the common shared area of the code that should be well maintained. 

Thank you for this great PR!

pkg/serviceproxy/auth_provider.go
mikeshng: @kahirokunn this got merged but could you fix this when you have some cycles in a new PR? Thanks!

PR #349, opened 2026-08-05
:seedling: chore: prepare for v0.11.0 release

mikeshng:
The old v0.10.0 release is too old. Let's just get an emergency v0.11.0 release out. We can do another release when we feel needed.


kahirokunn:
Thank you 🙏

PR #351, opened 2026-08-08
:sparkles: fix(user-server): add liveness probe

kahirokunn:
The user-server already exposes an HTTP health endpoint on port 8000, but its Deployment does not configure Kubernetes to use it for liveness checks.


mikeshng (approved):
/lgtm

I am ok with separating this out. In case of anything it's easier to back out.

Thank you for your contributions as always!

PR #353, opened 2026-08-08
🐛 fix(proxyagent): render RBAC for additional authentication args

kahirokunn:
`additionalServiceProxyArgs` can enable OIDC without setting `oidcIssuerURL`. The proxy then accepted credentials, but the chart omitted users/groups impersonation RBAC, causing managed-cluster requests to fail with 403.

This renders the rule for either configuration path. Configuring the same authentication option twice now fails during rendering instead of relying on argument order.

### DIFF
--- a/pkg/proxyagent/agent/agent_test.go
+++ b/pkg/proxyagent/agent/agent_test.go
@@ -1028,6 +1028,153 @@ func TestNewAgentAddon(t *testing.T) {
 	}
 }
 
+func TestImpersonationRBACConfiguration(t *testing.T) {
+	const (
+		addonName        = "open-cluster-management-cluster-proxy"
+		clusterName      = "cluster"
+		deployConfigName = "deploy-config"
+	)
+
+	tests := []struct {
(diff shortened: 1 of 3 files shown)


kahirokunn:
@mikeshng @tesshuflower Hi ✋ PTAL. Thank you 🙏

mikeshng (commented):
This whole helm helper file looks like something that is error prone, difficulty to understand and hard to maintain (need to sync with go code if there are changes). I am thinking it should be go code or limit the helm logic here some how.

@kahirokunn WDYT? I don't mind getting this in and then we revisit later. I think there are some improvements here besides this issue that I want to get in.

kahirokunn:
I was concerned that distributing logic between Helm and Go might reduce cohesion, but I think it's worth experimenting with. I also agree with merging this for now, as it will allow us to compare the before and after.

mikeshng (approved):
/lgtm

Merging for now as discussed. Should revisit later as discussed.

PR #354, opened 2026-08-12
:bug: reject malformed certificate blocks in CA bundles

kahirokunn:
Loading a CA bundle goes through `x509.CertPool.AppendCertsFromPEM`, which silently skips anything it cannot parse. A corrupted bundle is accepted as long as one certificate in it still parses (or, where the return value is ignored, even when nothing parses at all), and the process then quietly runs with partial trust, rejecting peers it should have accepted.

This switches the parsing to client-go's certutil, [the same code kube-apiserver uses when it reloads its client CA](https://github.com/kubernetes/kubernetes/blob/0f29094e5b73085e3802ecc1298ecae13866bfe6/staging/src/k8s.io/apiserver/pkg/server/dynamiccertificates/dynamic_cafile_content.go#L269-L280). A CERTIFICATE block that does not parse now fails startup.

- service-proxy: the root CA assembly moves into `loadRootCAs` so the boundary is unit-testable. The `--additional-service-ca` file is still optional: a missing file is only logged, and the config checker is still registered only when the file was read. A file that exists but contains no valid certificate now fails startup instead of being silently ignored.
- user-server: the service proxy CA is loaded with `certutil.NewPool`, which also rejects bundles where only some blocks parse.
- `pkg/util` (e2e helper): `buildTLSConfig` now rejects CA data without any valid certificate; previously an unparseable bundle produced an empty pool and a confusing handshake failure later.

Non-certificate PEM blocks are still ignored, and bytes that never form a PEM block (a truncated block, for example) are still skipped, same as kube-apiserver.

### DIFF
--- a/pkg/serviceproxy/service_proxy.go
+++ b/pkg/serviceproxy/service_proxy.go
@@ -15,6 +15,7 @@ import (
 	"github.com/spf13/cobra"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
+	certutil "k8s.io/client-go/util/cert"
 	"k8s.io/klog/v2"
 
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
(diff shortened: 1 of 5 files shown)


kahirokunn:

mikeshng (approved):
/lgtm

This is a super critical fix. We need this asap. FYI @tesshuflower 

Thank you again for your contribution!!!

PR #355, opened 2026-08-12
🌱 refactor(serviceproxy): restore auth provider dependency boundary

kahirokunn:
## Summary

Keep the authentication provider abstraction maintainable as provider-specific runtime dependencies evolve. Common orchestration should remain stable instead of gaining parameters for each provider requirement.

### DIFF
--- a/pkg/serviceproxy/auth_provider.go
+++ b/pkg/serviceproxy/auth_provider.go
@@ -65,10 +65,10 @@ func defaultAuthProviderFactories() []authProviderFactory {
 	}
 }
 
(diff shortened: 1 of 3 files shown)


kahirokunn:
@mikeshng Hi ✋ PTAL. Thank you 🙏

mikeshng (approved):
/lgtm

Yes, this is what I wanted from https://github.com/open-cluster-management-io/cluster-proxy/pull/348#discussion_r3769157500

Thank you for addressing it so quickly!

PR #356, opened 2026-08-14
🐛 fix(proxyserver): reconcile rendered resource changes after upgrades

kahirokunn:
## What changes

An addon-manager upgrade now brings the updatable resources rendered for the proxy server in line with that manager version, even when the ManagedProxyConfiguration has not changed. Changes to the Deployment, controller-owned in-cluster Service, RBAC, or generated Secret data therefore do not wait for an unrelated configuration or TLS update.

Previously, updates were gated by the ManagedProxyConfiguration generation and TLS configuration. A manager upgrade that changed its rendered output could leave existing resources on the old configuration indefinitely.

In-cluster Service updates preserve the API-assigned cluster IP and IP-family fields. The separately referenced LoadBalancer entrypoint Service keeps its existing create-only behavior.

## Verification

- go test ./pkg/...
- make test-integration
- kind upgrade validation on cpv-hub, cpv-spoke1, and cpv-spoke2: reconciled a stale Service render at the same configuration generation, preserved ClusterIP/IP-family fields, and kept all three cluster-proxy tunnels healthy

### DIFF
--- a/pkg/proxyserver/controllers/managedproxyconfiguration_controller.go
+++ b/pkg/proxyserver/controllers/managedproxyconfiguration_controller.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 	"time"
 
+	addonutils "open-cluster-management.io/addon-framework/pkg/utils"
 	proxyv1alpha1 "open-cluster-management.io/cluster-proxy/pkg/apis/proxy/v1alpha1"
 	"open-cluster-management.io/cluster-proxy/pkg/common"
 	"open-cluster-management.io/cluster-proxy/pkg/constant"
@@ -26,6 +27,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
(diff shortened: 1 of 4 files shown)


pkg/proxyserver/controllers/managedproxyconfiguration_controller.go
```diff
@@ -260,26 +270,29 @@ func (c *ManagedProxyConfigurationReconciler) ensure(incomingGeneration int64, g
...
-		return created, false, nil
-	}
+	// update if generation bumped, TLS config changed or the rendered manifest changed
+	tlsHashChanged := annotations[common.AnnotationKeyTLSConfigHash] !=
+		currentAnnotations[common.AnnotationKeyTLSConfigHash]
+	renderedHashChanged := renderedHash != currentAnnotations[common.AnnotationKeyRenderedHash]
+	if !created && (int(incomingGeneration) > currentGeneration || tlsHashChanged || renderedHashChanged) {
+		resourceToUpdate := resource
+		if gvk.Group == "" && gvk.Kind == "Service" {
+			resourceToUpdate, err = serviceForUpdate(resource, current)
```
mikeshng, line 288: Double check the rbac so it has permission to do this because before it didn't need to have the privilege to do this.

kahirokunn: Confirmed. This PR only adds update reconciliation for the controller-owned in-cluster Service, which is rendered into [config.Spec.ProxyServer.Namespace](https://github.com/open-cluster-management-io/cluster-proxy/blob/baa0fdce0b90b083d325a878be2306c93ed61a19/pkg/proxyserver/controllers/manifests.go#L57-L64). For the Helm-supported installation path, the chart sets that field to [.Release.Namespace](https://github.com/open-cluster-management-io/cluster-proxy/blob/baa0fdce0b90b083d325a878be2306c93ed61a19/charts/cluster-proxy/templates/managedproxyconfiguration.yaml#L13-L17).

The addon manager [runs as the cluster-proxy ServiceAccount](https://github.com/open-cluster-management-io/cluster-proxy/blob/baa0fdce0b90b083d325a878be2306c93ed61a19/charts/cluster-proxy/templates/manager-deployment.yaml#L20-L21), which is [bound to the namespaced addon-manager Role](https://github.com/open-cluster-management-io/cluster-proxy/blob/baa0fdce0b90b083d325a878be2306c93ed61a19/charts/cluster-proxy/templates/rolebinding.yaml#L6-L13). [That Role grants all verbs on core Services in the release namespace](https://github.com/open-cluster-management-io/cluster-proxy/blob/baa0fdce0b90b083d325a878be2306c93ed61a19/charts/cluster-proxy/templates/role.yaml#L3-L14), which includes update.

I verified the effective permission with:

    kubectl auth can-i update services \
      --as=system:serviceaccount:open-cluster-management-addon:cluster-proxy \
      --namespace=open-cluster-management-addon

This returns yes. I also exercised the actual Service update successfully in the kind upgrade test. The LoadBalancer entrypoint Service remains create-only, and this PR does not update Services outside the release namespace.

mikeshng (approved):
/lgtm 

Thank you for addressing my invalid concern!

Thanks again for all your contributions!

PR #357, opened 2026-08-14
🐛 fix(proxyserver): watch owned Secrets without duplicate cache

kahirokunn:
## Summary

Reconcile a `ManagedProxyConfiguration` when one of its owned certificate Secrets changes, using the existing Secret informer.

## Why

Certificate rotation writes new data to Secrets, while certificate conditions and expiry times in `ManagedProxyConfiguration.status` are read through the informer cache. A follow-up reconcile updates the status after the cache observes the new certificate.

Reusing the existing informer also avoids a second cluster-wide Secret cache. A Secret update does not restart proxy-server pods; a rollout still requires a change to the desired Deployment spec.


mikeshng (approved):
/lgtm

I was worried initially about watching the secret might cause too much reconciliation due to cert rotation or something like that but I think it should be ok in this case.

PR #367, opened 2026-08-25
:seedling: chore: dynamically fetch Go version in workflows

dhaiducek:
Currently the GitHub workflows rely on a static value for the Go version. It's more maintainable to specify that Go be fetched from the `go.mod` instead.


tesshuflower:
/lgtm

Thanks Dale this is nice!

PR #368, opened 2026-08-26
:bug: bug: fix user-server tunnel churn causing proxy-server saturation under load

tesshuflower:
## Problem

The user-server creates a new gRPC tunnel (full TCP+TLS negotiation) for
every incoming HTTP request, regardless of whether an existing tunnel to
the same cluster is idle. Under load this generates large numbers of
short-lived Proxy() handlers on the ANP proxy-server, causing:

- Lock contention on shared state (established map, PendingDial)
- Channel backpressure: `"Receive channel from agent/frontend is full"`
- Stream cancellations: `"Stream read from frontend cancelled"`
- Cleanup races: `"could not get frontend client for closing"`

### Commit 1: Cache one http.Transport per managed cluster

Replace the per-request transport creation with a cached transport per
cluster stored in a sync.Map. Go's http.Transport retains the underlying
net.Conn (the konnectivity tunnel) in its idle pool after each request
completes. Subsequent requests reuse the idle connection without creating
a new tunnel.

Transport pool uses hardcoded defaults appropriate for HTTP/2 multiplexing:
- maxIdleConnsPerHost=10
- maxConnsPerHost=10
- idleConnTimeout=90s

### Commit 2: Enable HTTP/2 and streaming flush on cached transport

With HTTP/2 enabled, concurrent REST API requests to the same cluster
multiplex over a single connection via HTTP/2 streams. The proxy-server
sees O(user-server replicas) Proxy() handlers per cluster regardless of
request concurrency, instead of one handler per concurrent request.

Upgrade requests (SPDY/WebSocket, e.g. kubectl exec, console sessions)
bypass the cached transport and use a dedicated HTTP/1.1 tunnel -- SPDY
compatibility is not affected.

If HTTP/2 negotiation fails, the transport falls back to HTTP/1.1
automatically, and MaxConnsPerHost acts as a safety net.

A `--enable-http2` flag (default true) is provided to disable HTTP/2 if
needed for compatibility.

`FlushInterval = -1` is set on the reverse proxy so that chunked/streaming
responses (e.g. Kubernetes watch requests using chunked transfer encoding)
are flushed to the client immediately instead of being buffered. Without
this, watch events sit in the reverse proxy buffer and downstream proxies
(e.g. HAProxy routes) close the connection after their idle timeout.

## Notes

- No changes to konnectivity client, ANP proxy-server, Helm charts, or CRDs
- The service-proxy already accepts HTTP/2 inbound (Go's http.Server enables
  it automatically with TLS)

### DIFF
--- a/charts/cluster-proxy/README.md
+++ b/charts/cluster-proxy/README.md
@@ -40,6 +40,7 @@ helm install cluster-proxy ./charts/cluster-proxy \
 | `enableImpersonation`                   | Grant permissions for hub token authentication                    | `true`                                          |
 | `featureGates.clusterProfile`           | Enable ClusterProfile integration                                | `false`                                         |
 | `userServer.enabled`                    | Generate and rotate the user-server serving certificate          | `false`                                         |
+| `userServer.enableHTTP2`                | Multiplex non-upgrade requests over reusable HTTP/2 tunnels       | `true`                                          |
 | `userServer.additionalSANs`             | Extra SANs for the generated user-server certificate             | `[]`                                            |
 | `exposedServicesConfigMapName`          | ConfigMap containing the service allowlist                        | `cluster-proxy-exposed-services`                 |
 | `exposedServices`                       | Services exposed through the service proxy path                   | `[]`                                            |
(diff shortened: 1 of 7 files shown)


mikeshng (commented):
This looks good to me.

/assign @kahirokunn

kahirokunn:
I opened tesshuflower/cluster-proxy#4 as a focused follow-up on top of this PR: https://github.com/tesshuflower/cluster-proxy/pull/4

It adds ManagedClusterAddOn-driven transport lifecycle and eviction, unknown-cluster 404s, shutdown cleanup, standard streaming behavior without global per-write flushes, Helm wiring for enable-http2, and real TLS/HTTP2/streaming/101-upgrade tests. I also validated API load, watch streaming, and kubectl exec against local kind-hub with kind-cluster1 and kind-cluster2, including HTTP/1.1 fallback, then restored the environment.

mikeshng (approved):
/lgtm

Thank you all!

